PR #11, opened 2018-05-12
Fixing cfg value source printing

tsiq-karold:
Fixes https://github.com/twosigma/webtau/issues/5


webtau-groovy/src/main/groovy/com/twosigma/webtau/cli/WebTauTestCliConfig.groovy
```diff
@@ -99,9 +105,10 @@ class WebTauTestCliConfig {
         }
     }
 
-    private def cliValue(String name, defaultValue) {
-        return commandLine.hasOption(name) ? commandLine.getOptionValue(name) :
-                defaultValue
+    private def setValueFromCli(ConfigValue configValue) {
```
MykolaGolubyev: should we use "try" or "ifPresent" to denote that it may not always set?

PR #15, opened 2018-05-14
Groovy script runner

MykolaGolubyev:


webtau-groovy/src/test/groovy/com/twosigma/webtau/cfg/WebTauGroovyCliArgsConfigHandlerTest.groovy
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+
+package com.twosigma.webtau.cfg
+
+import org.junit.Test
+
+class WebTauGroovyCliArgsConfigHandlerTest {
+    @Test
+    void "sets config essential config values during first run and overrides other cfg values during second"() {
```
tsiq-karold: "config essential config"?

webtau/src/main/java/com/twosigma/webtau/expectation/SystemTimerConfig.java
```diff
@@ -21,20 +21,22 @@
...
-        return cfg.waitTimeout();
+        return getCfg().waitTimeout();
     }
 
     @Override
     public long defaultTickMillis() {
         return 100;
     }
+
+    private WebTauConfig getCfg() {
```
tsiq-karold: `getCfg()` is sometimes static, sometimes not, feels like it should be consistent.

webtau/src/main/java/com/twosigma/webtau/http/config/WebTauHttpConfiguration.java
```diff
@@ -21,19 +21,21 @@
...
-        return HttpUrl.concat(cfg.getBaseUrl(), url);
+        return HttpUrl.concat(getCfg().getBaseUrl(), url);
     }
 
     @Override
     public HttpRequestHeader fullHeader(HttpRequestHeader given) {
         return given;
     }
+
+    private WebTauConfig getCfg() {
```
tsiq-karold: Should we rename `WebTauConfig.getInstance()` to `WebTauConfig.getCfg()` then you can use `getCfg()` everywhere with a static import without having to define an identical private `getCfg()` function everywhere.

Alternatively define a static public function `getCfg()` in some helper class so we can avoid the unnecessary duplication.

webtau/src/main/java/com/twosigma/webtau/expectation/SystemTimerConfig.java
MykolaGolubyev: Logic was to use static for classes with static methods, and non static otherwise. Your other suggestion is good. So I am renaming to getCfg.

webtau/src/main/java/com/twosigma/webtau/WebTauDsl.java
```diff
@@ -55,7 +53,7 @@
     }
 
     public static WebTauConfig getCfg() {
-        return WebTauConfig.getInstance();
+        return WebTauConfig.getCfg();
```
tsiq-karold: Do you still need this method here?

MykolaGolubyev: yes, this one is important, that's how you can do
`cfg.myVarFromConfig` in your tests. I.e. officially exposed cfg access through a single Dsl class

PR #19, opened 2018-05-16
Return non-0 code if tests fail or error

tsiq-karold:
Fixes https://github.com/twosigma/webtau/issues/18


webtau-groovy/src/main/groovy/com/twosigma/webtau/cli/WebTauCliApp.groovy
```diff
@@ -166,6 +171,8 @@ class WebTauCliApp implements StandaloneTestListener {
         }
 
         tests.add(test)
+
+        failure = failure || test.hasError() || test.failed
```
MykolaGolubyev: do we need to non 0 exit in case of skipped tests?
there is another handler afterAllTests, wonder if that is more appropriate place to make a final decision (you have more data there)

PR #26, opened 2018-05-20
initial junit 5 runner extension

MykolaGolubyev:
no documentation since maven repo is not setup yet


webtau-junit5/src/main/java/com/twosigma/webtau/junit5/Webtau.java
```diff
@@ -14,8 +14,19 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.ElementType.ANNOTATION_TYPE;
+
+@Target({ TYPE, ANNOTATION_TYPE })
```
tsiq-karold: Why ANNOTATION_TYPE?

webtau-junit5/src/main/java/com/twosigma/webtau/junit5/WebtauJunitExtension.java
```diff
@@ -0,0 +1,95 @@
...
+        StepReporters.remove(test);
+
+        ReportTestEntry reportTestEntry = test.getReportTestEntry();
+        reportTestEntry.setRan(true);
+        ReportJavaTestEntries.add(reportTestEntry);
+
+        TestResultPayloadExtractors.extract(reportTestEntry.getSteps().stream())
+                .forEach(reportTestEntry::addTestResultPayload);
+
+        ShutdownHook.INSTANCE.noOp();
```
tsiq-karold: Not sure I understand the benefit of this noOp invocation.  Is it just to force registration of the shutdown hook?  Do we need to actually have our ShutdownHook class to do that?

webtau-report/src/main/java/com/twosigma/webtau/report/ReportTestEntry.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static com.twosigma.webtau.cfg.WebTauConfig.getCfg;
+import static com.twosigma.webtau.reporter.TestStatus.*;
+
+public class ReportTestEntry {
```
tsiq-karold: If only Java had data classes 😄

webtau-junit5/src/main/java/com/twosigma/webtau/junit5/Webtau.java
MykolaGolubyev: You could have your own Extension say @IntegrationTests and annotate it wit @WebTau to inherit additional listeners. Strictly speaking we don't need to support this scenario, but that's the general idea of Junit 5 extensions (my current understanding)

webtau-junit5/src/main/java/com/twosigma/webtau/junit5/WebtauJunitExtension.java
MykolaGolubyev: I basically want to register hook only once, and only when any webtau test was called first. Wanted to avoid fields set and checks.

webtau-report/src/main/java/com/twosigma/webtau/report/ReportTestEntry.java
MykolaGolubyev: Is this Kotlin plug?

webtau-junit5/src/main/java/com/twosigma/webtau/junit5/WebtauJunitExtension.java
tsiq-karold: Could you not register it in `WebtauJunitExtension`'s static initialiser block?  Not sure which of these options I dislike the least 😄

webtau-report/src/main/java/com/twosigma/webtau/report/ReportTestEntry.java
tsiq-karold: I guess but I'm not sure it makes sense to introduce a new language just to help with one POJO!

webtau-junit5/src/main/java/com/twosigma/webtau/junit5/WebtauJunitExtension.java
MykolaGolubyev: I think in this case it will register hook even if the actual test is not called. Since Extension will be referred and therefore initialized during test files scan. I may be wrong.

tsiq-karold: Nope, you're quite right.

PR #33, opened 2018-05-24
capture http call related errors

MykolaGolubyev:
Report is now able to show/distinct http calls with error (e.g. wrong url protocol).
TestStep refactoring.

### DIFF
--- a/webtau-core/src/main/java/com/twosigma/webtau/reporter/TestStep.java
+++ b/webtau-core/src/main/java/com/twosigma/webtau/reporter/TestStep.java
@@ -16,38 +16,49 @@
 
 package com.twosigma.webtau.reporter;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
 
 import static com.twosigma.webtau.reporter.stacktrace.StackTraceUtils.renderStackTrace;
 import static java.util.stream.Collectors.toList;
 
-public class TestStep<E> {
-    private E context;
+public class TestStep<C, R> {
+    private C context;
 
     private TokenizedMessage inProgressMessage;
     private Supplier<TokenizedMessage> completionMessageSupplier;
-    private Supplier action;
+    private Supplier<R> action;
     private TokenizedMessage completionMessage;
 
     private boolean isInProgress;
     private boolean isSuccessful;
 
-    private List<TestStep<?>> children;
-    private TestStep<?> parent;
+    private List<TestStep<?, ?>> children;
+    private TestStep<?, ?> parent;
     private String stackTrace;
 
     private List<TestStepPayload> payloads;
 
-    private static ThreadLocal<TestStep<?>> currentStep = new ThreadLocal<>();
+    private static ThreadLocal<TestStep<?, ?>> currentStep = new ThreadLocal<>();
 
-    public static <E> TestStep<E> create(E context,
+    public static <C> TestStep<C, Void> create(C context,
+                                               TokenizedMessage inProgressMessage,
+                                               Supplier<TokenizedMessage> completionMessageSupplier,
+                                               Runnable action) {
+        return create(context, inProgressMessage, completionMessageSupplier, toSupplier(action));
+    }
+
+    public static <C, R> TestStep<C, R> create(C context,
                                          TokenizedMessage inProgressMessage,
                                          Supplier<TokenizedMessage> completionMessageSupplier,
-                                         Supplier action) {
-        TestStep<E> step = new TestStep<>(context, inProgressMessage, completionMessageSupplier, action);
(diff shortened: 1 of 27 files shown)


webtau-core/src/main/java/com/twosigma/webtau/reporter/TestStep.java
```diff
@@ -45,7 +45,7 @@
     public static <E> TestStep<E> create(E context,
                                          TokenizedMessage inProgressMessage,
                                          Supplier<TokenizedMessage> completionMessageSupplier,
-                                         Runnable action) {
+                                         Supplier action) {
```
tsiq-karold: Presumably this supplier always supplies an R?  If so, couldn't we parameterise TestStep on it?  Not really fond of using generic classes with no generic params (not least because IntelliJ moans at you for it 😄 )

webtau-core/src/main/java/com/twosigma/webtau/reporter/ValueMatcherExpectationSteps.java
```diff
@@ -29,34 +31,35 @@
     public static <C, V> void shouldStep(C context, V value, StepReportOptions stepReportOptions, TokenizedMessage valueDescription, ValueMatcher valueMatcher) {
         executeStep(context, value, valueDescription, valueMatcher, false,
                 tokenizedMessage(action("expecting")),
-                () -> actual(value).should(valueMatcher), stepReportOptions);
+                runnable(() -> actual(value).should(valueMatcher)), stepReportOptions);
```
tsiq-karold: Looks like all invocations of `executeStep` take what is actually a `Runnable`.  Therefore, I think we should either have `executeStep` take a `Runnable` and convert to a `Supplier` (rather than doing the conversion in each `executeStep` invocation site) or we create an overload of `TestStep.create` which takes a `Runnable` and converts it.

webtau-core/src/main/java/com/twosigma/webtau/reporter/ValueMatcherExpectationSteps.java
```diff
@@ -69,4 +72,11 @@
 
         step.execute(stepReportOptions);
     }
+
+    private static Supplier runnable(Runnable r) {
```
tsiq-karold: This could return `Supplier<Void>`

webtau-core/src/main/java/com/twosigma/webtau/reporter/stacktrace/StackTraceUtils.java
```diff
@@ -113,11 +113,9 @@ private static String filterStackTrace(Throwable t, Predicate<String> filter) {
         String stackTrace = renderStackTrace(t);
         List<String> lines = Arrays.asList(stackTrace.split("\n"));
 
-        return lines.get(0) + "\n" +
-                lines.subList(1, lines.size())
-                        .stream()
-                        .filter(filter)
-                        .collect(Collectors.joining("\n"));
+        return lines.stream()
```
tsiq-karold: Why this change?  I assumed we had the lines.get(0) because we didn't want to apply filtration on the message part of the stack trace and now that's no longer the case.  It is of course possible my assumption was incorrect...

webtau-core/src/main/java/com/twosigma/webtau/reporter/ValueMatcherExpectationSteps.java
tsiq-karold: Having looked through the rest of the review, I'd go for the `TestStep.create` overload as there are other classes that wrap a `Runnable` into a `Supplier` in exactly the same way.

webtau-core/src/main/java/com/twosigma/webtau/reporter/TestStep.java
MykolaGolubyev: I thought to add it, the problem is the actual type we only now at the `step.execute()` time and not the create time.

```
ThisIsTheType id = http.post(...) { 
    return id
}
```

so we can define it as a generic, but it won't do any good. Of course I may be missing some generic magic knowledge.

webtau-core/src/main/java/com/twosigma/webtau/reporter/ValueMatcherExpectationSteps.java
MykolaGolubyev: Makes sense. Thanks.

webtau-core/src/main/java/com/twosigma/webtau/reporter/stacktrace/StackTraceUtils.java
MykolaGolubyev: For one liners it was doing extra '\n'. Then I looked at what filters are being passed. We always check now for the presence of `at` part, which never happens on first line.
This filter function is private and we control use cases. That was justification.

tsiq-karold: Makes sense

PR #41, opened 2018-06-10
initial DateAndStringCompareToHandler to compare date/time and string

MykolaGolubyev:


MykolaGolubyev:
I can never get timezones right from the first attempt. Please review carefully.

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/DateAndStringCompareToHandler.java
```diff
@@ -0,0 +1,168 @@
...
+
+        void compare() {
+            if (actual instanceof LocalDate && expected instanceof LocalDate) {
+                compareLocalDates((LocalDate) actual, (LocalDate) expected);
+            } else if (actual instanceof ZonedDateTime && expected instanceof LocalDate) {
+                compareZonedDateTimeAndLocalDate((ZonedDateTime) actual, (LocalDate) expected);
+            } else if (actual instanceof ZonedDateTime && expected instanceof ZonedDateTime) {
+                compareZonedDateTimes((ZonedDateTime) actual, (ZonedDateTime) expected);
+            } else {
+                throw new UnsupportedOperationException("combination is not supported:\n" +
```
tsiq-karold: Why not actual LocalDate with expected ZonedDateTime?

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/DateAndStringCompareToHandler.java
```diff
@@ -0,0 +1,168 @@
...
+
+            for (FormatParser parser: parsers) {
+                try {
+                    return parser.convert(actualAsText);
+                } catch (DateTimeParseException ignored) {
+                }
+            }
+
+            throw new UnsupportedOperationException("cannot parse " + actualAsText + "\navailable formats:\n" +
+                    parsers.stream().map(FormatParser::toString).collect(Collectors.joining("\n")));
```
tsiq-karold: Instead of all this, could do `String.join('\n', parsers)`

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/DateAndStringCompareToHandler.java
MykolaGolubyev: got mentally drained by this time :) do you want to give it a shot?
Plus I didn't have a use case to be sufficiently motivated.

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/DateAndStringCompareToHandler.java
MykolaGolubyev: keep forgetting about this one! thanks.

MykolaGolubyev: Hmm, String.join expects `Iterable<? extends CharSequence>`. `parsers` is `List<FormatParser>`

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/DateAndStringCompareToHandler.java
tsiq-karold: Hmm, yeah, what time portion it should use is a good question.  Let's park it until we have a usecase.

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/DateAndStringCompareToHandler.java
tsiq-karold: Oh, yeah, this always annoyed me and I always forget.  On many occasions, I've ended up writing my own version of `String.join` which takes essentially an Iterable of objects.  At least where I couldn't use Eclipse Collections (or Kotlin) which handle this problem natively.

Issue #24, opened 2018-05-18
Matcher enhancements: <, >, <=, >= and date compatibility


tsiq-karold:
I would like to be able to do something like:
```
body.someDate.should <= LocalDate.now()
```
Or something similar.

MykolaGolubyev:
added `beGreatearThan`, `beLessThan`, `beGreaterThanOrEqual`, `beLessThanOrEqual` for now.
Handling of `<=` requires AST and AST is not implemented yet, even for ==.

PR #42, opened 2018-06-11
data node support for greater/less comparisons

MykolaGolubyev:
add CompareToComparator modes instead of negative flag
add documentation matchers page


MykolaGolubyev:
this in combination with #41 closes #24

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/CompareToComparator.java
```diff
@@ -44,126 +53,62 @@
...
+        return new CompareToComparator(null);
     }
 
-    /**
-     * creates new instance of comparator without any collected mismatches. Preserves isNegative switch.
-     * @return fresh copy of equal comparator
-     */
-    public CompareToComparator freshCopy() {
-        return new CompareToComparator(isNegative);
+    public static CompareToComparator comparator(AssertionMode assertionMode) {
```
tsiq-karold: Why did you choose static methods rather than c'tors?

webtau-rest-groovy/src/test/groovy/com/twosigma/webtau/http/HttpTest.groovy
```diff
@@ -236,6 +243,52 @@ class HttpTest {
         assert transformed == []
     }
 
+    @Test
+    void "equality matcher"() {
+        http.get("/end-point") {
+            id.shouldNot == 0
```
tsiq-karold: Is it worth adding dates to these tests?

MykolaGolubyev: I wonder if a separate dates section should be there. I didn't since dates were in a separate PR. This PR is dates agnostic. I will keep updating docs in a separate PR.

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/CompareToComparator.java
MykolaGolubyev: It made more sense when there were two different versions.

tsiq-karold: OK.  We probably should change to c'tors at some point but not a biggie.

webtau-rest-groovy/src/test/groovy/com/twosigma/webtau/http/HttpTest.groovy
tsiq-karold: OK.

PR #45, opened 2018-06-12
URL to operation rename

### LINKED ISSUE #44: Coverage info in report should say operations not urls
The coverage is based on operations (i.e. path + method), not urls.

tsiq-karold:
In test report, coverage info is labelled as url whereas it's really operation coverage.

Fixes #44


webtau-reactjs/src/report/OverallSummary.css
```diff
@@ -22,7 +22,7 @@
 }
 
 .overall-http-calls-time,
-.overall-http-calls-coverage {
+.overall-operations-coverage {
```
MykolaGolubyev: seems like http calls is till valid here? 
need some indication what operations are as this report is also used for UI testing.

webtau-reactjs/src/report/OverallSummary.js
```diff
@@ -25,7 +25,7 @@ function OverallSummary({report, onSwitchToHttpCalls, onSwitchToSkippedHttpCalls
     return (
         <div className="overall-summary">
             <CallsTiming report={report} onSwitchToHttpCalls={onSwitchToHttpCalls}/>
-            <UrlCoverageSummary report={report} onSwitchToSkippedHttpCalls={onSwitchToSkippedHttpCalls}/>
+            <OperationCoverageSummary report={report} onSwitchToSkippedHttpCalls={onSwitchToSkippedHttpCalls}/>
```
MykolaGolubyev: same as previous css comment

webtau-reactjs/src/report/OverallSummary.css
tsiq-karold: Yeah, I guess this one probably still make sense as http-calls, will put it back.

webtau-reactjs/src/report/OverallSummary.js
tsiq-karold: Not sure this one is the same really.  UrlCoverageSummary presumably doesn't get used in a UI report?  Would be weird to have this be called UrlCoverageSummary when all the things it contains are "operation something".

MykolaGolubyev: I mean the word `operation` is too generic. Maybe `OpenApi` prefix?
In UI `operation` could be button click and coverage will be percentage of buttons clicked.

tsiq-karold: I see.  I misunderstood.  Will update.

PR #49, opened 2018-06-14
add ability to compose matchers

MykolaGolubyev:


webtau-core/src/main/java/com/twosigma/webtau/data/render/ToStringRenderer.java
```diff
@@ -19,6 +19,8 @@
 public class ToStringRenderer implements DataRenderer {
     @Override
     public String render(Object data) {
-        return data.toString();
+        return data instanceof String ?
```
tsiq-karold: The functionality of this renderer has changed a bit I'd say. It's not really ToStringRenderer but just StringRenderer.

webtau-core/src/main/java/com/twosigma/webtau/data/render/DataRenderers.java
```diff
@@ -40,6 +40,7 @@ public static String render(Object data) {
         renders.addAll(ServiceUtils.discover(DataRenderer.class));
         renders.add(new TableDataRenderer());
         renders.add(new ToStringRenderer());
+        renders.add(new AnyRenderer());
```
tsiq-karold: One could argue you don't need to actually have an `AnyRenderer` class and just add `Object::toString` to the list but I think that's a stylistic choice more than anything else so happy to leave that up to you.

tsiq-karold: Or `Objects::toString` if you want the null safety though it's not required with the first renderer being the `NullRenderer`.

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/GreaterThanMatcher.java
```diff
@@ -74,4 +74,9 @@ public boolean negativeMatches(ActualPath actualPath, Object actual) {
         compareToComparator = CompareToComparator.comparator();
         return compareToComparator.compareIsLessOrEqual(actualPath, actual, expected);
     }
+
+    @Override
+    public String toString() {
+        return GreaterLessEqualMatcherRenderer.render(this, compareToComparator, expected);
```
tsiq-karold: Some static function imports for things like `render` and `comparator()` might actually make the code nicer but it's also a stylistic choice so up to you.

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/NotEqualMatcher.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Override
+    public String negativeMatchedMessage(ActualPath actualPath, Object actual) {
+        return "equals to " + DataRenderers.render(expected) + "\n" +
+                comparator.generateEqualMatchReport();
+    }
+
+    @Override
+    public String negativeMismatchedMessage(ActualPath actualPath, Object actual) {
+        return "doesn't equal to " + DataRenderers.render(expected) + ", but should\n" +
```
tsiq-karold: "doesn't equal to" sounds a bit weird to me, I think should be either "isn't equal to" or "doesn't equal".  Maybe run it by someone who is an actual native English speaker though, I'm just an approximation to that 😄

webtau-core/src/test/groovy/com/twosigma/webtau/expectation/code/ThrowExceptionMatcherGroovyTest.groovy
```diff
@@ -20,94 +20,95 @@ import org.junit.Rule
...
 
     @Test
     void "should validate exception class and expected message pattern"() {
-        thrown.expectMessage("\nmismatches:\n" +
-                "\n" +
-                "expected exception message:    actual string: error message\n" +
-                "                            expected pattern: error \\d\n" +
-                "expected exception class:   actual: class java.lang.IllegalArgumentException <java.lang.Class>\n" +
-                "                          expected: class java.lang.UnsupportedOperationException <java.lang.Class>")
+        thrown.expectMessage('\nmismatches:\n' +
```
tsiq-karold: What's with all the `"` to `'` changes?

webtau-rest-groovy/src/test/groovy/com/twosigma/webtau/http/HttpTest.groovy
```diff
@@ -313,6 +321,28 @@ class HttpTest {
         http.doc.capture("end-point-dates-matchers")
     }
 
+    @Test
+    void "matchers combo"() {
+        http.get("/end-point-mixed") {
+            list.should contain(lessThanOrEqual(2)) // lessThanOrEqual will be matched against each value
```
tsiq-karold: Just to make sure I understand correctly, this is conceptually equivalent to:
```
list.stream().any(x -> x <=2)
```
rather than `all()`?

webtau-core/src/main/java/com/twosigma/webtau/data/render/ToStringRenderer.java
MykolaGolubyev: good catch. Renaming

webtau-core/src/main/java/com/twosigma/webtau/data/render/DataRenderers.java
MykolaGolubyev: will try to use that instead

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/GreaterThanMatcher.java
MykolaGolubyev: when I static import outside of DSL I prefer method itself to be less ambiguous. Any suggestion for the class name itself?

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/NotEqualMatcher.java
MykolaGolubyev: what about `equal to` above? I am all to make it english as long as it is symmetric.

webtau-core/src/test/groovy/com/twosigma/webtau/expectation/code/ThrowExceptionMatcherGroovyTest.groovy
MykolaGolubyev: I need to use " inside and also want to switch to using `'` by default.

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/NotEqualMatcher.java
tsiq-karold: I would say either "is equal to" or "equals"

webtau-rest-groovy/src/test/groovy/com/twosigma/webtau/http/HttpTest.groovy
MykolaGolubyev: yes. should the function names be changed to make it less confusing?

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/NotEqualMatcher.java
tsiq-karold: But I'm not really a native speaker so maybe get another opinion.

webtau-rest-groovy/src/test/groovy/com/twosigma/webtau/http/HttpTest.groovy
tsiq-karold: It's not that it was confusing.  The actual behaviour is what I would have inferred from the names, just wanted to make sure it definitely wasn't the other one.

Only alternative I can think of would be for the lessThanOrEqual and friends which are intended to be used in composition should be anElementLessThanOrEqual?  But that'll only make sense when composed with a contain not something else...

PR #51, opened 2018-06-16
groovy ast transformation to replace a.should==e with a.should equal(e)

### LINKED ISSUE #3: Groovy AST to handle should syntax on maps
`myMap.should == 'wrongvalue` gives no op. 
It is not an issue when assertion is done in the context of `http.get|post`. But if the same syntax is used outside that it just gives no errors. 

Need to create Groovy AST to replace calls like `myVar.should ==` with `ActualValue.value('myVar', myVar).should(equal(..))`.

Need to replace existing runtime way of doing it with exception "should not reach here" inside `Should` and `ShouldNot` classes

MykolaGolubyev:
this also closes #3


MykolaGolubyev:
@tsiq-karold I am figuring out the build error, but if you have time - please proceed with review. AST is also a blocker for proper byte[] response comparison

PR #52, opened 2018-06-18
add binary response support

MykolaGolubyev:


webtau-rest-groovy/src/test/groovy/com/twosigma/webtau/http/HttpTest.groovy
```diff
@@ -268,6 +270,18 @@ class HttpTest {
...
+    @Test
+    void "binary content"() {
+        byte[] expectedImage = ResourceUtils.binaryContent("dummy-image.png")
+
+        byte[] content = http.get("/binary") {
+            body.should == expectedImage
+            return body
+        }
+
+        content.length.should == expectedImage.length
```
tsiq-karold: Curious why you have this outside the assertions block passed to `http.get`?

Is it purely to test that returning binary from the get works and your average webtau user should just do the length check in the get assertions?

MykolaGolubyev: Yes, it is to test that it gets returned back. 
This snippet will not be used in docs as an example. I suppose I could use the same assertion as inside the block.

PR #54, opened 2018-06-18
add extra params to header provider

MykolaGolubyev:


MykolaGolubyev:
If you need a different set of auth headers for different urls/servers. Same idea as `baseUrl` impl, where you essentially can router your test to different servers.

PR #61, opened 2018-06-22
Fix null serialization

tsiq-karold:


tsiq-karold:
Don't merge this just yet, still doing some testing.

tsiq-karold:
@MykolaGolubyev can you re-review please?  I've ran all my tests which use webtau against this and they all passed.

webtau-rest/src/main/java/com/twosigma/webtau/http/Http.java
```diff
@@ -514,6 +513,10 @@ private static DataNode createBodyDataNode(HttpResponse response) {
         }
     }
 
+    private static boolean nullOrEmptyResponse(String content) {
+            return content == null || content.isEmpty();
+        }
```
MykolaGolubyev: alignment issue

webtau-rest/src/main/java/com/twosigma/webtau/http/HttpDocumentation.java
```diff
@@ -38,14 +38,19 @@ public void capture(String artifactName) {
...
+                && !lastValidationResult.getRequestContent().isEmpty()) {
             String fileName = "request." + fileExtensionForType(lastValidationResult.getRequestType());
             FileUtils.writeTextContent(path.resolve(fileName),
                     prettyPrintContent(lastValidationResult.getRequestType(),
                             lastValidationResult.getRequestContent()));
         }
 
-        if (lastValidationResult.getResponseType() != null) {
+        if (lastValidationResult.getResponseType() != null
+                && lastValidationResult.getResponseTextContent() != null
```
MykolaGolubyev: since it is a 3rd time we need it, could we please add a method to StringUtils similar to apache utils to check for null or empty at once.

webtau-utils/src/main/java/com/twosigma/webtau/utils/StringUtils.java
```diff
@@ -80,6 +80,14 @@ public static String concatWithIndentation(String prefix, String multilineText)
                 "");
     }
 
+    public static boolean nullOrEmpty(String s) {
+        return s == null || s.trim().isEmpty();
```
MykolaGolubyev: I see optimization you have done here. We do change behavior slightly. 
Maybe we should distinct empty vs blank? 
I wonder if someone can actually send "          " back and it will be swept out by our new logic.

PR #63, opened 2018-06-24
http operation overloads for empty body and no validation

MykolaGolubyev:


webtau-rest-groovy/src/main/groovy/com/twosigma/webtau/http/HttpExtensions.groovy
```diff
@@ -63,10 +71,18 @@ class HttpExtensions {
         return http.put(url, new JsonRequestBody(requestBody), closureToHttpResponseValidator(validation))
     }
 
+    static def put(Http http, String url, Closure validation) {
```
tsiq-karold: Not sure put with empty body really makes sense?  Post is often used with an empty body for action like resources, things which aren't CRUD but not sure put really is.

webtau-rest-groovy/src/test/groovy/com/twosigma/webtau/http/HttpTest.groovy
```diff
@@ -274,6 +276,61 @@ class HttpTest {
         assert varArgHeader == mapBasedHeader
     }
 
+    @Test
+    void "no body request"() {
+        def noBodyExpectation = {
+            body.should == ""
```
tsiq-karold: Should it?  I would have thought no body should make body be null, not empty string.

tsiq-karold: Oh, I see.  This is really an artifact of how `TestServerResponseEcho` is implemented.  You can ignore this question :)

webtau-rest-groovy/src/main/groovy/com/twosigma/webtau/http/HttpExtensions.groovy
MykolaGolubyev: Good point. I will leave for now since someone can just not use semantic REST.

PR #66, opened 2018-06-29
Pass query params to API spec validator

tsiq-karold:


webtau-rest/src/test/groovy/com/twosigma/webtau/http/HttpUrlTest.groovy
```diff
@@ -66,4 +67,28 @@ class HttpUrlTest {
...
+        def expectedNullParams = [a: [null], b: ['c']]
+
+        HttpUrl.extractQueryParams('relative/path?a=b&c=d').should == expectedParams
+        HttpUrl.extractQueryParams('https://localhost:8080/relative/path?a=b&c=d').should == expectedParams
+        HttpUrl.extractQueryParams('https://localhost:8080/relative/path?a=b&a=c').should == expectedMultiParams
+        HttpUrl.extractQueryParams('https://localhost:8080/relative/path?a[0]=b&a[1]=c').should == expectedMultiParams
+
+        HttpUrl.extractQueryParams('https://localhost:8080/relative/path?a&b=c').should == expectedNullParams
+
+        Assert.assertNull(HttpUrl.extractQueryParams('relative/path'))
```
tsiq-karold: blah.should == null didn't work in this case.  I will do a separate PR to fix that but for now assertNull it is.

webtau-rest/src/main/java/com/twosigma/webtau/http/HttpUrl.java
```diff
@@ -42,6 +46,34 @@ public static String extractPath(String url) {
         }
     }
 
+    public static Map<String, List<String>> extractQueryParams(String url) {
```
MykolaGolubyev: can we make it return empty map instead of null?

webtau-rest/src/main/java/com/twosigma/webtau/http/HttpUrl.java
```diff
@@ -42,6 +46,34 @@ public static String extractPath(String url) {
...
+        String queryParams;
+        if (!isFull(url)) {
+            queryParams = extractQueryParamsFromRelativeUrl(url);
```
MykolaGolubyev: most likely subjective, but what about declaring and assigning variable in one statement?

webtau-rest/src/main/java/com/twosigma/webtau/http/HttpUrl.java
tsiq-karold: Yes

webtau-rest/src/main/java/com/twosigma/webtau/http/HttpUrl.java
tsiq-karold: You mean with a tertiary operator?

MykolaGolubyev: yes. hence subjective :) no Kotlin if-else option unfortunately. 
Another option is to extract to a function and do if-else-return. 
I have fear of unassigned variable after C++.

tsiq-karold: I do too but with a local var like here, the compiler will throw an error for unassigned var usage.

webtau-rest/src/test/groovy/com/twosigma/webtau/http/HttpUrlTest.groovy
```diff
@@ -66,4 +66,28 @@ class HttpUrlTest {
             HttpUrl.extractPath('garbage://localhost:8080/relative/path').should == '/relative/path'
         } should throwException(IllegalArgumentException, 'invalid url: garbage://localhost:8080/relative/path')
     }
+
+    @Test
+    void "extract query params from a given url"() {
+        def expectedParams = [a: ['b'], c: ['d']]
```
MykolaGolubyev: since extract params is a generic util function, do we want values to be defaulted to arrays?
when we do `http.get`, values you pass there are not arrays.

tsiq-karold: Presumably you can pass them as arrays too?  How else do you send them for query params which can be repeated?

I could make it return a map of string to string or list but it'll make all the code that handles it much more horrible.

MykolaGolubyev: I would pass them as an array when it is an array.
Maybe symmetry is not important here.

PR #88, opened 2018-07-12
Allow validating just request or response

tsiq-karold:


webtau-open-api/src/main/java/com/twosigma/webtau/openapi/OpenApiSpecValidator.java
```diff
@@ -56,7 +54,23 @@ public void validateApiSpec(HttpValidationResult result) {
         SimpleRequest request = buildRequest(result);
         SimpleResponse response = buildResponse(result);
 
-        ValidationReport validationReport = openApiValidator.validate(request, response);
+        ValidationReport validationReport;
+        switch (validationMode) {
```
MykolaGolubyev: small nit pick: could extract as a function so assign+break is replaced with single return

PR #102, opened 2018-07-18
fix exit code regression. add test.

MykolaGolubyev:


MykolaGolubyev:
something weird going on here. Test passes in IDE, but fails in maven.

PR #103, opened 2018-07-19
Maven run mojo

MykolaGolubyev:
Fixes https://github.com/twosigma/webtau/issues/99


webtau-groovy/src/main/groovy/com/twosigma/webtau/cli/WebTauCliApp.groovy
```diff
@@ -114,7 +112,10 @@ class WebTauCliApp implements StandaloneTestListener, ReportGenerator {
     }
 
     private List<Path> testFiles() {
-        return cliConfigHandler.testFiles.collect { Paths.get(it) }
+        return cliConfigHandler.testFiles.collect { fileName ->
+            def path = Paths.get(fileName)
+            return path.isAbsolute() ? path : cfg.workingDir.resolve(path)
```
tsiq-karold: Thinking out loud, not convinced it's a good idea but... is it worth having a helper function on `WebTauConfig` to do this absolute/relative thing?  It's not the only place you do the exact same tertiary check.

webtau-maven-plugin-test/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>com.twosigma.webtau</groupId>
+                <artifactId>webtau-maven-plugin</artifactId>
+                <version>${project.version}</version>
+                <configuration>
+                    <workingDir>${project.basedir}/src/tests/groovy</workingDir>
```
tsiq-karold: So just to make sure I understand this correctly, to use non-default env you'd do `<env>myEnv</env>`?

MykolaGolubyev: yes

webtau-groovy/src/main/groovy/com/twosigma/webtau/cli/WebTauCliApp.groovy
MykolaGolubyev: Was thinking the same. Makes sense to me.

PR #112, opened 2018-07-23
add ansi color config to toggle

MykolaGolubyev:
Can't rely on System.console() == null since it is the case when running
from IDE. There is no easy way to check in java if output is being
redirected. So instead relying on a explicit flag


motlin:
jansi manages to enable/disable color codes automatically, only falling back on config as a last resort.

https://github.com/fusesource/jansi/blob/master/jansi/src/main/java/org/fusesource/jansi/AnsiConsole.java

MykolaGolubyev:
@motlin thanks for the comment! 
if I remember correctly, they use jni to do so.

PR #114, opened 2018-07-24
--verbose param to control output's verbosity 

MykolaGolubyev:


MykolaGolubyev:
will be adding more cases, possibly one at a time

webtau-config/src/main/java/com/twosigma/webtau/cfg/WebTauConfig.java
```diff
@@ -44,6 +44,8 @@
     private final ConfigValue config = declare("config", "config path", () -> "webtau.cfg");
     private final ConfigValue env = declare("env", "environment id", () -> "local");
     private final ConfigValue url = declare("url", "base url for application under test", NO_DEFAULT);
+    private final ConfigValue verbose = declare("verbose", "output verbosity level. " +
```
tsiq-karold: I don't think -verbose often takes a param, it just means run in verbose mode.  Maybe we can have either different params (c.f. Maven) or call it `logLevel` or `verbosityLevel` or something like that?

PR #123, opened 2018-08-07
add response headers to be available for validation

MykolaGolubyev:


webtau-http/src/main/java/com/twosigma/webtau/http/HttpDataNodes.java
```diff
@@ -0,0 +1,115 @@
...
+import com.twosigma.webtau.utils.JsonParseException;
+import com.twosigma.webtau.utils.JsonUtils;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+
+class HttpDataNodes {
+    private static final Function<String, Object> asIs = (v) -> v;
```
tsiq-karold: Isn't this `Function.identity()`?  Or do you specifically need one which returns Object?

tsiq-karold: Never mind, looking at the usage you probably couldn't use Function.identity().

PR #138, opened 2018-08-20
Provide a better error message when headers are null

tsiq-karold:


webtau-http/src/main/java/com/twosigma/webtau/http/Http.java
```diff
@@ -535,6 +535,10 @@ private HttpResponse request(String method, String fullUrl,
             connection.setRequestMethod(method);
             connection.setRequestProperty("Content-Type", requestBody.type());
             connection.setRequestProperty("Accept", requestBody.type());
+
+            if (requestHeader == null) {
+                throw new IllegalArgumentException("Request header is null, check your header provider is not returning null");
+            }
```
MykolaGolubyev: could you please put an empty line here to separate logical blocks?
also wonder if we could spare the whole url connection creation.

webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpTest.groovy
```diff
@@ -875,6 +876,15 @@ class HttpTest implements HttpConfiguration {
         http.lastValidationResult.errorMessage.should == ~/java.lang.ClassCastException: .*cannot be cast to .*HttpURLConnection/
     }
 
+    @Test
+    void "provides appropriate error when headers are null"() {
+        code {
+            http.get("/end-point", (HttpRequestHeader) null, (HttpResponseValidator) {})
+        } should throwException(HttpException, ~/error during http\.get/)
+
+        http.lastValidationResult.errorMessage.should == ~/java.lang.IllegalArgumentException: Request header is null/
```
MykolaGolubyev: sweet!

webtau-http/src/main/java/com/twosigma/webtau/http/Http.java
tsiq-karold: Done.

MykolaGolubyev: what about "moving it to the top" part to avoid creating connection in case of bad header?

PR #143, opened 2018-08-23
header node properties access is case insensitive

MykolaGolubyev:


webtau-http/src/main/java/com/twosigma/webtau/http/validation/HeaderDataNode.java
```diff
@@ -93,4 +99,12 @@ public String toString() {
     public DataNode statusCode() {
         return dataNode.get("statusCode");
     }
+
+    private String findMatchingCaseInsensitiveKey(String name) {
+        String lowerCaseName = name.toLowerCase();
+        return dataNode.asMap().keySet().stream()
+                .filter(k -> k.toLowerCase().equals(lowerCaseName))
+                .findFirst()
+                .orElse("");
```
tsiq-karold: Why not return the optional from `findFirst` and then instead of `key.isEmpty()` use `optional.isPresent()`?

tsiq-karold:
Didn't mean to approve, meant to click comment!

tsiq-karold:
Btw, why not store them lower case in the map in the first place when you get them from UrlHttpConneiton (or whatever that class is called, I forget)?

webtau-http/src/main/java/com/twosigma/webtau/http/validation/HeaderDataNode.java
tsiq-nick: old habits. let me change that.

PR #159, opened 2018-09-10
Maven plugin to respect skipTests

tsiq-karold:
This is based on the surefire plugin behaviour: http://maven.apache.org/plugins-archives/maven-surefire-plugin-2.12.4/examples/skipping-test.html and code: https://github.com/apache/maven-surefire/blob/7149edc6f24fa8bff06372e24a177e86d4960d8c/maven-surefire-common/src/main/java/org/apache/maven/plugin/surefire/AbstractSurefireMojo.java#L177

It started off as just adding the skip tests capability and a test for it but found a couple of issues I had to fix:
1. the maven plugin test was not actually executing the tests because it requires the execution part to be set up.  I think there is a way to make that work by default but I don't remember off the top of my head what that is
1. adding a fix for the above unearthed a bug in the OpenAPISpec loading which resulted in an exception being thrown if no spec was configured in the config file.  This is because we resolved working dir with `cfg.spec.asString()` (pseudocode).  This by default returns `""` and therefore the spec was the working dir which then threw an exception

### DIFF
--- a/webtau-maven-plugin-test-skip/pom.xml
+++ b/webtau-maven-plugin-test-skip/pom.xml
@@ -24,7 +24,7 @@
         <version>0.30-SNAPSHOT</version>
     </parent>
 
-    <artifactId>webtau-maven-plugin-test</artifactId>
+    <artifactId>webtau-maven-plugin-test-skip</artifactId>
 
     <dependencies>
         <dependency>
@@ -49,6 +49,7 @@
                     </execution>
                 </executions>
                 <configuration>
+                    <skipTests>true</skipTests>
                     <workingDir>${project.basedir}/src/tests/groovy</workingDir>
                     <tests>
                         <directory>${project.basedir}/src/tests/groovy</directory>
--- a/webtau-maven-plugin/src/main/groovy/com/twosigma/webtau/maven/WebTauMavenRunner.groovy
+++ b/webtau-maven-plugin/src/main/groovy/com/twosigma/webtau/maven/WebTauMavenRunner.groovy
@@ -50,20 +50,24 @@ class WebTauMavenRunner extends AbstractMojo {
         if (skipTests()) {
             getLog().info("Skipping webtau tests")
         } else {
-            def fileSetManager = new FileSetManager()
-            def files = fileSetManager.getIncludedFiles(tests) as List
+            runTests()
+        }
+    }
 
-            getLog().info("test files:\n    " + files.join("\n    "))
+    private void runTests() {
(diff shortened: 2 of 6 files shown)


webtau-maven-plugin-test-skip/pom.xml
```diff
@@ -0,0 +1,64 @@
...
+                <executions>
+                    <execution>
+                        <phase>test</phase>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <skipTests>true</skipTests>
```
tsiq-karold: Just for the record, running `mvn test` on this module without this line resulted in a failure as expected.

webtau-maven-plugin/src/main/groovy/com/twosigma/webtau/maven/WebTauMavenRunner.groovy
```diff
@@ -39,21 +39,31 @@ class WebTauMavenRunner extends AbstractMojo {
...
+    protected boolean skip
+
     @Override
     void execute() throws MojoExecutionException, MojoFailureException {
-        def fileSetManager = new FileSetManager()
-        def files = fileSetManager.getIncludedFiles(tests) as List
+        if (skipTests()) {
+            getLog().info("Skipping webtau tests")
+        } else {
+            def fileSetManager = new FileSetManager()
```
MykolaGolubyev: will it make sense to extract `else` branch now so `if-else` is short and easy to follow?

webtau-open-api/src/main/java/com/twosigma/webtau/openapi/OpenApiSpecConfig.java
```diff
@@ -26,12 +26,16 @@
 
 public class OpenApiSpecConfig implements WebTauConfigHandler {
     static final ConfigValue specUrl = declare("openApiSpecUrl",
-            "url of OpenAPI 2 spec against which to validate http calls", () -> "");
+            "url of OpenAPI 2 spec against which to validate http calls", () -> null);
```
MykolaGolubyev: a bit confused. `null` when used in `getAsString` will still return empty string. 
Could you please explain how does it fix the problem?

tsiq-karold: That’s my bad.  I initially had a different fix where I was using `getAsObject` but then changed the approach and forgot to roll this bit back.  I’ll fix

PR #170, opened 2018-09-28
add watch for auto test run on file(s) change

MykolaGolubyev:


webtau-groovy/src/main/groovy/com/twosigma/webtau/cli/interactive/FileWatcher.groovy
```diff
@@ -0,0 +1,143 @@
...
+        rootPath.traverse type: FileType.DIRECTORIES, visit: {
+            register(it)
+        }
+    }
+
+    private static WatchService createWatchService() {
+        return FileSystems.getDefault().newWatchService()
+    }
+
+    static void main(String[] args) {
```
tsiq-karold: Guessing this was for testing?  Probably should remove it.

PR #180, opened 2018-10-16
Java/matcher examples

MykolaGolubyev:


webtau-core-groovy/src/test/groovy/com/twosigma/webtau/data/expectation/equality/handlers/GroovyNullCompareToHandlerTest.groovy
```diff
@@ -39,4 +39,23 @@ class GroovyNullCompareToHandlerTest {
             actual(NullObject.getNullObject()).shouldNot(equal(null))
         } should throwException(AssertionError)
     }
+
+    @Test
+    void todo() {
```
tsiq-karold: todo?

webtau-core/src/main/java/com/twosigma/webtau/Ddjt.java
```diff
@@ -69,6 +71,10 @@ public static ContainMatcher containing(Object expected) {
         return new ContainMatcher(expected);
     }
 
+    public static <K, V> Map<K, V> createMap(Object... kvs) {
```
tsiq-karold: Would it be better to take a vararg of pairs (or map entries)?  We could define a static helper so that a user would do:
```
createMap(
    entry("foo", "bar"),
    entry("hello", "world")
)
```
We can then get compile time type checks instead of throwing a class cast exception and the runtime `% 2` check also goes away.

webtau-docs/webtau/REST/matchers.md
```diff
@@ -5,86 +5,122 @@ type: two-sides
...
+myValue.shouldBe > 10
+```
 
-Additionally `shouldBe` and `shouldNotBe` alias keywords are available to make certain matcher combinations easier to read
- 
-```groovy
+Java:
+```java
+myValue.should(contain(10))
+myValue.shouldNot(equal("hello"))
```
tsiq-karold: `myValue.shouldBe greaterThan(10)` does not look like it would compile in Java 😬

webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
```diff
@@ -921,15 +868,15 @@ class HttpTest implements HttpConfiguration {
...
     private static void assertStatusCodeMismatchRegistered() {
         http.lastValidationResult.mismatches.should contain(~/statusCode/)
     }
 
     private static byte[] binaryFile(String path) {
         return [1, 2, 3] as byte[]
     }
+
+    private static Path testResourcePath(String relativePath) {
+        return Paths.get("../webtau-http/src/test/resources/image.png")
```
tsiq-karold: How is the build succeeding with this?  You're using `testResourcePath` in all the tests but ignoring its parameter so it's always doing a get for this png.

tsiq-karold: Oh, I see, it's always passing in the same png param which happens to match what you actually do in this method.  Should fix this though, either use the param or don't ask for param and call the method `testImagePath()`.

webtau-http/src/test/java/com/twosigma/webtau/http/HttpJavaTest.java
```diff
@@ -0,0 +1,181 @@
...
+            body.get("amount").should(equal(30));
+
+            body.get("list").should(equal(Arrays.asList(1, 2, 3)));
+
+            body.get("object").get("k1").should(equal(
+                    Pattern.compile("v\\d"))); // regular expression matching
+
+            body.get("object").should(equal(createMap("k1", "v1", "k3", "v3"))); // matching only specified fields and can be nested multiple times
+
+            body.get("complexList").should(equal(header("k1" , "k2").values( // matching only specified fields, but number of entries must be exact
```
tsiq-karold: Not sure I get what this header().values() is doing.  Can you explain?

webtau-http/src/test/java/com/twosigma/webtau/http/HttpJavaTest.java
```diff
@@ -0,0 +1,181 @@
...
+                                                        "v1" ,  30,
+                                                        "v11",  40)));
+        });
+
+        http.doc.capture("end-point-object-equality-matchers");
```
tsiq-karold: `HttpGroovyTest` and `HttpTest` seem to be using same doc capture names, won't they override each other?  Or are they being run with different doc roots?

webtau-http/src/test/java/com/twosigma/webtau/http/HttpJavaTest.java
```diff
@@ -0,0 +1,181 @@
...
+            body.get("list").get(1).shouldNotBe(greaterThan(2));
+        });
+
+        http.doc.capture("end-point-numbers-matchers");
+    }
+
+    @Test
+    public void containMatcher() {
+        http.get("/end-point-list", (header, body) -> {
+            body.should(contain(createMap("k1", "v1", "k2", "v2")));
```
tsiq-karold: Maybe this would read better if `createMap` was renamed to just `map`?

webtau-http/src/test/java/com/twosigma/webtau/http/HttpJavaTest.java
```diff
@@ -0,0 +1,181 @@
...
+import static com.twosigma.webtau.utils.CollectionUtils.createMap;
+
+public class HttpJavaTest implements HttpConfiguration  {
+    private static final HttpTestDataServer testServer = new HttpTestDataServer();
+
+    private static final int PORT = 7824;
+
+    @BeforeClass
+    public static void startServer() {
+        testServer.start(PORT);
```
tsiq-karold: At some point we probably should change this to start on a random port and return the port number otherwise we'll inevitably get some clashes.

webtau-core-groovy/src/test/groovy/com/twosigma/webtau/data/expectation/equality/handlers/GroovyNullCompareToHandlerTest.groovy
MykolaGolubyev: o boy :). thanks.

webtau-core/src/main/java/com/twosigma/webtau/Ddjt.java
MykolaGolubyev: fair. Initially it was a private method to help with impl details. Entry could be to generic for a static import though. Java doesn't have the Kotlin context magic. And it is being for testing only.

webtau-docs/webtau/REST/matchers.md
MykolaGolubyev: Was lazy and didn't move it to test files initially. Moving now.

webtau-http/src/test/java/com/twosigma/webtau/http/HttpJavaTest.java
MykolaGolubyev: same thought occurred. `map` is a functional paradigm as well and may be confusing. Give me more ideas please.

webtau-http/src/test/java/com/twosigma/webtau/http/HttpJavaTest.java
MykolaGolubyev: yeah... Is there an easy way that is built in in a jetty that I am using?

tsiq-karold: Pretty sure Jetty allows you to specify port 0 and then it'll pick a random port.  Its server class should have a `port()` or `httpPort()` method (can't remember which).

tsiq-karold: I think with Jetty you may have to actually get the connectors from the server and then their port rather than asking the server itself but in any case it's definitely doable.

webtau-http/src/test/java/com/twosigma/webtau/http/HttpJavaTest.java
tsiq-karold: `aMapOf()`?

webtau-http/src/test/java/com/twosigma/webtau/http/HttpJavaTest.java
MykolaGolubyev: it is Java way to define table data. Maybe should rename header to tableHeader?

webtau-http/src/test/java/com/twosigma/webtau/http/HttpJavaTest.java
MykolaGolubyev: if they do it shouldn't matter, idea is tests should be identical in what they assert.

webtau-http/src/test/java/com/twosigma/webtau/http/HttpJavaTest.java
MykolaGolubyev: https://github.com/twosigma/webtau/issues/181

webtau-http/src/test/java/com/twosigma/webtau/http/HttpJavaTest.java
tsiq-karold: Ah, yes.  I assumed it was something to do with http headers.

PR #188, opened 2018-10-24
Run test servers on random free ports

### LINKED ISSUE #181: auto port assignment for tests that run as part of travis
Tests like `HttpGroovyTest` hardcode port number for testing. Need to replace with auto port discovery.

tsiq-karold:
Fixes #181

### DIFF
--- a/webtau-feature-testing/pom.xml
+++ b/webtau-feature-testing/pom.xml
@@ -103,6 +103,25 @@
             </plugin>
 
             <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>build-helper-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>reserve-springboot-port</id>
+                        <goals>
+                            <goal>reserve-network-port</goal>
+                        </goals>
+                        <phase>process-resources</phase>
+                        <configuration>
+                            <portNames>
+                                <portName>springboot.http.port</portName>
+                            </portNames>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+
+            <plugin>
                 <groupId>org.springframework.boot</groupId>
                 <artifactId>spring-boot-maven-plugin</artifactId>
                 <configuration>
@@ -115,6 +134,11 @@
                         <goals>
                             <goal>start</goal>
                         </goals>
+                        <configuration>
+                            <arguments>
+                                <argument>--server.port=${springboot.http.port}</argument>
+                            </arguments>
+                        </configuration>
                     </execution>
                     <execution>
                         <id>stop-app</id>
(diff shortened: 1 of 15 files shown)


webtau-feature-testing/pom.xml
```diff
@@ -102,6 +102,25 @@
                 </configuration>
             </plugin>
 
+            <plugin>
```
tsiq-karold: This is apparently the way to get a random free port with the springboot maven plugin: https://docs.spring.io/spring-boot/docs/current/maven-plugin/examples/it-random-port.html

webtau-testapp/src/main/java/com/example/demo/springboot/app/SpringBootDemoApp.java
```diff
@@ -17,7 +18,9 @@
 @Import({springfox.documentation.spring.data.rest.configuration.SpringDataRestConfiguration.class})
 public class SpringBootDemoApp {
     public static void main(String[] args) {
-        SpringApplication.run(SpringBootDemoApp.class, args);
+        ConfigurableApplicationContext context = SpringApplication.run(SpringBootDemoApp.class, args);
+        String port = context.getEnvironment().getProperty("local.server.port");
+        System.out.println("Springboot test server running on port " + port);
```
MykolaGolubyev: do we still need this line?

webtau-junit4-examples/src/test/java/com/example/tests/junit4/CustomerCrudSingleJavaTest.java
```diff
@@ -14,6 +15,12 @@
     private Map<String, Object> customerPayload = createCustomerPayload();
     private Map<String, Object> changedCustomerPayload = createChangedCustomerPayload();
 
+    @BeforeClass
+    public static void setUpClass() {
+        String port = System.getProperty("springboot.http.port", "8080");
```
MykolaGolubyev: do we still need 8080 here?

tsiq-karold: I added the default value to make debugging in an IDE easier because the port randomisation happens in Maven.  If you want to run the tests in an IDE, you'd just start the springboot app which will default to 8080 so it's useful to then just run these tests without having to set a system prop.

webtau-testapp/src/main/java/com/example/demo/springboot/app/SpringBootDemoApp.java
tsiq-karold: That's debatable.  The Maven plugin for reserving the port logs the number so it's probably not needed.  I'll remove it.

webtau-junit4-examples/src/test/java/com/example/tests/junit4/CustomerCrudSingleJavaTest.java
MykolaGolubyev: ah, I didn't realize it is a second parameter. somehow just saw it as the only value

PR #198, opened 2018-11-07
skip tests based on condition

MykolaGolubyev:


webtau-groovy/src/main/groovy/com/twosigma/webtau/cfg/ConfigFileEnvironmentsCollector.groovy
```diff
@@ -31,7 +31,7 @@ class ConfigFileEnvironmentsCollector {
         this.workingDir = workingDir
         this.configPath = configPath
         this.groovy = GroovyRunner.createWithCustomScriptClass(workingDir, IgnoreAllConfigScriptClass.class)
-        this.environments = []
+        this.environments = ['local']
```
tsiq-karold: Do we need this?

webtau-groovy/src/main/groovy/com/twosigma/webtau/cli/WebTauCliApp.groovy
```diff
@@ -198,7 +198,7 @@ class WebTauCliApp implements StandaloneTestListener, ReportGenerator {
     @Override
     void generate(Report report) {
         def summary = report.createSummary()
-        problemCount = (int) (summary.failed + summary.errored + summary.skipped)
+        problemCount = (int) (summary.failed + summary.errored)
```
tsiq-karold: Hmmm, there was benefit to including skipped scenarios in the problem count as it would catch someone accidentally pushing a `sscenario`.  Any thoughts on way to reintroduce that capability?

I understand why this change was required here and I'm not too keen on introducing another result type so wondering whether we can do something else that would prevent pushing `sscenario`s.

webtau-standalone-runner-groovy/src/main/groovy/com/twosigma/webtau/runner/standalone/StandaloneTestRunner.groovy
```diff
@@ -74,13 +80,42 @@ class StandaloneTestRunner {
...
+        }
     }
 
     void sscenario(String description, Closure code) {
+        handleDisabledByCondition(description, code) { test ->
+            exclusiveTests.add(test)
+        }
+    }
+
+    void dscenario(String description, Closure code) {
```
tsiq-karold: So `dscenario` is the equivalent of `@Ignore`, right?

PR #201, opened 2018-11-20
Produce accurate expected messages for comparisons other than equality

tsiq-karold:
Currently, this test:
```groovy
scenario('should not be null assertion') {
    def value = null
    value.shouldNot == null
}
```

Results in this exception:
```
java.lang.AssertionError: 
equals [null], but shouldn't
mismatches:

[value]:   actual: null
         expected: null
```

The `expected: null` part is very confusing as we're in fact expecting it to be not null.  This PR fixes the rendering of the expected value for not equals, less than, etc.

I created a new test which exercises all the `CompareToHandler`s because some had their own rendering.  There are a few however that are missing:
* RecordAndMapCompareTo
* TableDataCompareTo
* TraceableValueCompareTo
* ValueMatcherCompareTo

I can add them in due course but I'll need to figure out exactly how to test them.

### DIFF
--- a/webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/HandlerMessages.java
+++ b/webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/HandlerMessages.java
@@ -16,22 +16,55 @@
 
 package com.twosigma.webtau.expectation.equality.handlers;
 
+import com.twosigma.webtau.expectation.equality.CompareToComparator;
+
 import static com.twosigma.webtau.utils.TraceUtils.renderValueAndType;
 
 class HandlerMessages {
     private static final String ACTUAL_PREFIX = "  actual: ";
     private static final String EXPECTED_PREFIX = "expected: ";
 
-    static String renderActualExpected(Object actual, Object expected) {
+    static String renderActualExpected(CompareToComparator.AssertionMode assertionMode, Object actual, Object expected) {
         return ACTUAL_PREFIX + renderValueAndType(actual) + "\n" +
-                EXPECTED_PREFIX + renderValueAndType(expected);
+                expected(assertionMode, renderValueAndType(expected));
     }
 
-    static String renderActualExpected(Object convertedActual, Object convertedExpected,
+    static String renderActualExpected(CompareToComparator.AssertionMode assertionMode,
+                                       Object convertedActual, Object convertedExpected,
                                        Object actual, Object expected) {
         return ACTUAL_PREFIX + renderValueAndType(convertedActual) +
                 "(before conversion: " + renderValueAndType(actual) + ")\n" +
-                EXPECTED_PREFIX + renderValueAndType(convertedExpected) +
-                "(before conversion: " + renderValueAndType(expected) + ")";
+                expected(assertionMode, renderValueAndType(convertedExpected) +
+                "(before conversion: " + renderValueAndType(expected) + ")");
+    }
+
+    static String expected(CompareToComparator.AssertionMode assertionMode, Object expected) {
+        return expected(EXPECTED_PREFIX, assertionMode, expected);
+    }
+
+    static String expected(String prefix, CompareToComparator.AssertionMode assertionMode, Object expected) {
+        String expectedMsg = prefix;
+
+        switch (assertionMode) {
+            case NOT_EQUAL:
+                expectedMsg += "not ";
+                break;
+            case GREATER_THAN:
+                expectedMsg += "> ";
+                break;
+            case GREATER_THAN_OR_EQUAL:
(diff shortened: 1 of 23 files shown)


tsiq-karold:
@MykolaGolubyev any thoughts on this?

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/HandlerMessages.java
```diff
@@ -16,22 +16,55 @@
...
-                "(before conversion: " + renderValueAndType(expected) + ")";
+                expected(assertionMode, renderValueAndType(convertedExpected) +
+                "(before conversion: " + renderValueAndType(expected) + ")");
+    }
+
+    static String expected(CompareToComparator.AssertionMode assertionMode, Object expected) {
+        return expected(EXPECTED_PREFIX, assertionMode, expected);
+    }
+
+    static String expected(String prefix, CompareToComparator.AssertionMode assertionMode, Object expected) {
```
MykolaGolubyev: nice

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/MapAndBeanCompareToHandler.java
```diff
@@ -39,11 +39,7 @@ private boolean isMapOfProps(Object o) {
     }
 
     private boolean isBean(Object o) {
-        if (o instanceof Iterable || o instanceof Map) {
-            return false;
-        }
-
-        return true;
+        return !(o instanceof Iterable || o instanceof Map);
```
MykolaGolubyev: duh :)

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/MapAndBeanCompareToHandler.java
```diff
@@ -55,10 +51,10 @@ public void compareEqualOnly(CompareToComparator comparator, ActualPath actualPa
         expectedMap.keySet().forEach(p -> {
             ActualPath propertyPath = actualPath.property(p);
 
-            if (! actualAsMap.containsKey(p)) {
-                comparator.reportMissing(this, propertyPath, expectedMap.get(p));
-            } else {
+            if (actualAsMap.containsKey(p)) {
```
MykolaGolubyev: ouch, that looked like a real bug :(

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/HandlerMessages.java
```diff
@@ -16,22 +16,55 @@
...
+        String expectedMsg = prefix;
+
+        switch (assertionMode) {
+            case NOT_EQUAL:
```
MykolaGolubyev: minor nit to move this out as a function to use `returns` and avoid extra spaces at the end.

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/MapAndBeanCompareToHandler.java
tsiq-karold: No, I just flipped the if statement.  I can't remember why though.  I think just because I felt it made more sense this way round 😄

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/HandlerMessages.java
tsiq-karold: I'll move it into a function, it's nicer than this large inline switch.  However, still need to add spaces somewhere and can't just prepend the expected text with a space for the equality case.

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/HandlerMessages.java
```diff
@@ -43,28 +47,11 @@ static String expected(CompareToComparator.AssertionMode assertionMode, Object e
...
-                expectedMsg += "less than or equal to ";
-                break;
-        }
-
-        expectedMsg += expected;
-        expectedMsg += "\n";
-        return expectedMsg;
+        return Stream.of(prefix, assertionMode.getMessage(), expected)
+                .map(Objects::toString)
+                .map(String::trim)
```
MykolaGolubyev: why trim?

tsiq-karold: Because various places pass in a prefix which already ends in a space.

PR #204, opened 2018-11-28
Fix test failure with Java 9

tsiq-karold:
This is due to https://issues.apache.org/jira/browse/GROOVY-8338.  Instead of upgrading Groovy, I have made the code uglier but did create an issue to remember to upgrade: https://github.com/twosigma/webtau/issues/203

### DIFF
--- a/webtau-core/src/test/groovy/com/twosigma/webtau/expectation/equality/handlers/HandlerMessagesTest.groovy
+++ b/webtau-core/src/test/groovy/com/twosigma/webtau/expectation/equality/handlers/HandlerMessagesTest.groovy
@@ -221,14 +221,14 @@ class HandlerMessagesTest {
     @Test
     void "equal stream and iterable"() {
         code {
-            actual(Stream.of(1)).should(equal(Arrays.asList(2)))
+            actual(Arrays.asList(1).stream()).should(equal(Arrays.asList(2)))
         } should throwException(AssertionError, ~/expected: 2/)
     }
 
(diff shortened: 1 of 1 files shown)


webtau-core/src/test/groovy/com/twosigma/webtau/expectation/equality/handlers/HandlerMessagesTest.groovy
```diff
@@ -221,14 +221,14 @@ class HandlerMessagesTest {
     @Test
     void "equal stream and iterable"() {
         code {
-            actual(Stream.of(1)).should(equal(Arrays.asList(2)))
+            actual(Arrays.asList(1).stream()).should(equal(Arrays.asList(2)))
```
tsiq-nick: [1].stream()?

tsiq-nick: Stream.of(1)?

tsiq-karold: Stream.of(1) is what was causing the problem in the first place!

tsiq-nick: O boy :) clearly need more decaf green tea

PR #212, opened 2018-12-03
Support gzip encoding

tsiq-karold:


webtau-http/src/main/java/com/twosigma/webtau/http/HttpDataNodes.java
```diff
@@ -70,6 +70,7 @@ private void populateHeaderData() {
         addCamelCaseVersion("Location", "location", asIs);
         addCamelCaseVersion("Content-Location", "contentLocation", asIs);
         addCamelCaseVersion("Content-Length", "contentLength", Integer::valueOf);
+        addCamelCaseVersion("Content-Encoding", "contentEncoding", asIs);
```
tsiq-karold: I will refactor/rename/tidy-up this class in a subsequent PR.  Didn't want to muddle this PR with unrelated changes.

webtau-http/src/test/java/com/twosigma/webtau/http/HttpJavaTest.java
```diff
@@ -81,24 +82,28 @@ public void matchersBasicExample() {
     @Test
     public void equalityMatcher() {
         http.get("/end-point", (header, body) -> {
-            body.get("id").shouldNot(equal(0));
-            body.get("amount").should(equal(30));
+            endPointExpectations(body);
```
MykolaGolubyev: don't we use these in docs?

tsiq-karold: ah, good point.  I'll put this back.

PR #213, opened 2018-12-03
Header shortcuts are generated in a case insensitive manner

tsiq-karold:
While doing this I also got rid of HttpDataNodes which was a) badly named and b) not really necessary.


webtau-http/src/main/java/com/twosigma/webtau/http/validation/HeaderDataNode.java
```diff
@@ -17,19 +17,52 @@
...
+        headerData.put("contentType", response.getContentType());
+
+        response.getHeader().forEachProperty(headerData::put);
+
+        translations.forEach(translation -> addCamelCaseVersion(headerData, translation));
+
+        this.dataNode = DataNodeBuilder.fromMap(new DataNodeId("header"), headerData);
+    }
+
+    private static void addCamelCaseVersion(Map<String, Object> headerData, CamelCaseTranslation translation) {
```
tsiq-karold: I realise that doing this for each header we're trying to translate isn't terribly efficient, essentially O(n*m) where n is the number of translations and m the number of headers.  However, for any realistic number of headers and translations this is not going to be an issue at all.  If we feel it would, I can implement this in a different way but it would be a little less obvious.

PR #226, opened 2018-12-11
do not force cast config value to string

MykolaGolubyev:


webtau-config/src/main/java/com/twosigma/webtau/cfg/WebTauConfig.java
```diff
@@ -105,12 +105,13 @@ public ConfigValue findConfigValue(String key) {
...
 
-    public String get(String key) {
+    @SuppressWarnings("unchecked")
+    public <E> E get(String key) {
         Stream<ConfigValue> allValues =
                 Stream.concat(enumeratedCfgValues.values().stream(), freeFormCfgValues.stream());
 
         Optional<ConfigValue> configValue = allValues.filter(v -> v.match(key)).findFirst();
-        return configValue.map(ConfigValue::getAsString).orElse("");
+        return (E) configValue.map(ConfigValue::getAsObject).orElse("");
```
tsiq-karold: The default in `orElse` won't really work with the cast to `E`.

MykolaGolubyev: ouch, thanks. need to understand what is the implication if I return null.

tsiq-karold: Unfortunately, based on my experience with this stuff, you're likely to get some NPEs :( but I'd be happy to be proven wrong

MykolaGolubyev: passed most likely because we have default values set for string config params. And we haven't tested it in the actual test codebase.

PR #228, opened 2018-12-15
expose get method to extract data node underlying value

MykolaGolubyev:


MykolaGolubyev:
Ouch, you are right. Let's chat later to figure out if it should be reverted.

PR #242, opened 2019-01-02
Verbose DSL scenario names

tsiq-karold:
A few changes in this PR:
- adding `singleScenario` and `disabledScenario` as aliases for `sscenario` and `dscenario` respectively.  This is to aid readability but still allow conciseness if desired
- adding `dscenario` to documentation.  I highjacked and renamed the selective skipping chapter for this.  Feel free to disagree with that move
- fixing a bug with `dscenario` and test runners where it would invoke the scenario anyway.  Should check what happens in interactive mode though because I think we should execute it if someone runs the disabled scenario explicitly but not if someone runs the entire file

### DIFF
--- a/webtau-feature-testing/examples/scenarios/concept/runOnlySelected.groovy
+++ b/webtau-feature-testing/examples/scenarios/concept/runOnlySelected.groovy
@@ -12,7 +12,11 @@ sscenario('step two') {
     // test that you want to focus on
 }
 
-scenario('step three') {
+singleScenario('step three') {
+    // test that you want to focus on
+}
+
+scenario('step four') {
     http.put('/extra') {
         // ...
     }
--- a/webtau-groovy/src/main/groovy/com/twosigma/webtau/WebTauGroovyDsl.groovy
+++ b/webtau-groovy/src/main/groovy/com/twosigma/webtau/WebTauGroovyDsl.groovy
@@ -51,6 +51,10 @@ class WebTauGroovyDsl extends WebTauDsl {
         }
     }
 
+    static void singleScenario(String description, Closure code) {
+        sscenario(description, code)
+    }
+
     static void sscenario(String description, Closure code) {
         if (!testRunner) {
             runAdHoc(description, code)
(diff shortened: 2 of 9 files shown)


webtau-feature-testing/examples/scenarios/concept/runOnlySelected.groovy
```diff
@@ -12,8 +12,12 @@ sscenario('step two') {
     // test that you want to focus on
 }
 
-scenario('step three') {
+singleScenario('step three') {
```
MykolaGolubyev: the reason for `sscenario` name was not really because it was `single`. I "borrowed" it from Jasmine, where `it` was a test declaration, and `iit` was run only this.
I like `singleScenario` but not sure how it translates when there are multiple `single` scenarios?

webtau-groovy/src/main/groovy/com/twosigma/webtau/WebTauGroovyDsl.groovy
```diff
@@ -59,11 +63,15 @@ class WebTauGroovyDsl extends WebTauDsl {
...
+    static void disabledScenario(String description, Closure code) {
+        dscenario(description, code)
+    }
+
     static void dscenario(String description, Closure code) {
         if (!testRunner) {
             ConsoleOutputs.out(Color.YELLOW, "disabled test ", Color.GREEN, FontStyle.BOLD, description)
         } else {
-            testRunner.scenario(description, code)
+            testRunner.dscenario(description, code)
```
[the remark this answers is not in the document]
MykolaGolubyev: good catch

MykolaGolubyev:
Thanks!

webtau-feature-testing/examples/scenarios/concept/runOnlySelected.groovy
tsiq-karold: Yeah, I wondered that too but struggled to come up with a better name.  I suppose technically it does not need to start with an `s` but I think it would be preferable.

tsiq-karold: Looks like Jasmine has moved away from that pattern to call these "focused" specs: https://jasmine.github.io/2.1/focused_specs.html.  Not sure I like that.

PR #243, opened 2019-01-02
Improve pattern rendering

tsiq-karold:
The problem this PR is addressing is demonstrated by an assertion such as:
```groovy
http.get("someurl") {
    field.should == ~/a regex pattern/
}
```

This results in an error:
```
failed expecting field to equal a regex pattern
```

If the pattern does not use any actual regex (and is therefore used to approximate a contains test), it's not really obvious that it's a pattern or where the actual pattern starts and stops.  Contrast that with the equivalent error for a string which would be `failed expecting field to equal "a string"` and the actual actual/expected rendering:
```
field:      actual string: the actual value
         expected pattern: a regex pattern
```
This makes it clear that it's a pattern test.

This PR results in the error becoming:
```
failed expecting field to equal pattern /a regex pattern/
```

### DIFF
--- a/webtau-groovy/src/main/groovy/com/twosigma/webtau/WebTauGroovyDsl.groovy
+++ b/webtau-groovy/src/main/groovy/com/twosigma/webtau/WebTauGroovyDsl.groovy
@@ -93,7 +93,7 @@ class WebTauGroovyDsl extends WebTauDsl {
      *
      * <pre>
      * def lazySharedData = createLazyResource("resource name") {
-     *      def result = callToInitializeTheResouce()
+     *      def result = callToInitializeTheResource()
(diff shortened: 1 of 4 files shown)


webtau-groovy/src/main/groovy/com/twosigma/webtau/WebTauGroovyDsl.groovy
```diff
@@ -93,7 +93,7 @@ class WebTauGroovyDsl extends WebTauDsl {
      *
      * <pre>
      * def lazySharedData = createLazyResource("resource name") {
-     *      def result = callToInitializeTheResouce()
+     *      def result = callToInitializeTheResource()
```
MykolaGolubyev: sauce :)

PR #248, opened 2019-01-07
set compare to handler

### LINKED ISSUE #209: Comparison of sets is ordered
It appears that comparing two java (hash)sets:

```set1.should == set2```

does not work correctly. The comparison is ordered elementwise - we ought to use set equality semantics here instead.

### PR DESCRIPTION
**Pull request by MykolaGolubyev**
Closes #209.

Set comparison complexity is going to be O(NM). In this PR set contains is not used and instead equality is checked by CompareToComparator to handle cases like

```
def actual = ['hello', 'world'] as Set
actual.should == [~/wo..d/, ~/h...o/] as Set
```

### DIFF
--- a/webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/SetCompareToHandler.java
+++ b/webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/SetCompareToHandler.java
@@ -57,17 +57,18 @@ public class SetCompareToHandler implements CompareToHandler {
         }
 
         void compare() {
-            expectedLeft.removeIf(this::checkActualAndRemoveIfMatch);
+            expectedLeft.removeIf(this::checkAndRemoveFromActualIfMatch);
 
             actualLeft.forEach(e -> comparator.reportExtra(SetCompareToHandler.this, actualPath, e));
             expectedLeft.forEach(e -> comparator.reportMissing(SetCompareToHandler.this, actualPath, e));
         }
 
-        private boolean checkActualAndRemoveIfMatch(Object expected) {
+        private boolean checkAndRemoveFromActualIfMatch(Object expected) {
             Iterator<?> it = actualLeft.iterator();
             int idx = 0;
             while (it.hasNext()) {
                 Object actual = it.next();
+
                 CompareToResult result = localComparator.compareUsingEqualOnly(actualPath, actual, expected);
                 if (result.isEqual()) {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/SetCompareToHandler.java [hunk 1]
```diff
@@ -0,0 +1,85 @@
...
+        Comparator(CompareToComparator compareToComparator, ActualPath actualPath, Set<?> actual, Set<?> expected) {
+            this.comparator = compareToComparator;
+            this.localComparator = CompareToComparator.comparator(compareToComparator.getAssertionMode());
+            this.actualPath = actualPath;
+            this.actualLeft = new ArrayList<>(actual);
+            this.expectedLeft = new ArrayList<>(expected);
+        }
+
+        void compare() {
+            expectedLeft.removeIf(this::checkActualAndRemoveIfMatch);
```
tsiq-karold: It's slightly confusing having the `removeIf` use a predicate with the word `removeIf` in it.  If I understand this correct, one removes from expected and the other from actual?  If that's the case, would a better name for the predicate be `checkAndRemoveFromActualIfMatch` or something along those lines?  Basically I think it would help to make it obvious from where the removal is happening in the predicate.

An alternative approach would be to do something like:
```
Map<Object, Object> matches = expectedLeft
                    .stream()
                    .map(this::getMatches)
                    .filter(Objects::nonNull)
                    .collect(Collectors.toMap(
                            Pair::getKey,
                            Pair::getValue
                    ));
            
            expectedLeft.removeAll(matches.keySet());
            actualLeft.removeAll(matches.values());
```
The `getMatches` is basically identical to `checkActualAndRemoveIfMatch` except it doesn't remove and instead of `true` returns `Pair.of(expected, actual)` and `null` instead of `false`.  That's probably not the best way to write that code but hopefully you get what I mean.  It's less efficient of course but arguably easier to understand.

MykolaGolubyev: I like `getMatches`, definitely more functional. I do not like maps returning empty and then filtering out bit though. 
I will proceed with rename since performance in case of sets is already a small concern.

tsiq-karold: Yeah, the empty filtration bothered me a bit too.

PR #271, opened 2019-02-17
open api schema validation as a separate test step

MykolaGolubyev:
I have no idea why it marks files as add and remove instead of a diff... trying again.
Tried a few other branches. Looks like rename in combination with large files changes makes git confused :(


MykolaGolubyev:
closing this PR and will try without rename

PR #273, opened 2019-02-20
make following redirects optional

tsiq-karold:
closes https://github.com/twosigma/webtau/issues/268

### DIFF
--- a/webtau-feature-testing/src/main/groovy/com/twosigma/webtau/featuretesting/WebTauTestRunner.groovy
+++ b/webtau-feature-testing/src/main/groovy/com/twosigma/webtau/featuretesting/WebTauTestRunner.groovy
@@ -71,6 +71,7 @@ class WebTauTestRunner implements StepReporter, StandaloneTestListener {
             args.add(testPath.toString())
 
             WebTauConfig.resetConfigHandlers()
+            getCfg().reset()
 
             def cliApp = new WebTauCliApp(args as String[])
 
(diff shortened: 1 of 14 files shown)


webtau-feature-testing/src/main/groovy/com/twosigma/webtau/featuretesting/WebTauTestRunner.groovy
```diff
@@ -71,6 +71,7 @@ class WebTauTestRunner implements StepReporter, StandaloneTestListener {
             args.add(testPath.toString())
 
             WebTauConfig.resetConfigHandlers()
+            getCfg().reset()
```
tsiq-karold: I had to add this because I found that in the features test, cfg was getting preserved between tests.  I think we've just gotten lucky before but in my case the disabled redirects test ran first and then passed, redirects on ran next but cfg still had it disabled.

webtau-utils/src/test/groovy/com/twosigma/webtau/utils/StringUtilsTest.groovy
```diff
@@ -77,4 +77,16 @@ line #_3\r""")
         def number = StringUtils.convertToNumber("123,000")
         assert number == 123000
     }
+
+    @Test
+    void "should start with"() {
+        Assert.assertEquals("/foo", StringUtils.ensureStartsWith("/foo", "/"))
```
MykolaGolubyev: should we use `assert` for consistency?

tsiq-karold: There's actually a mixture in this test but `assert` seems more prevalent so I've switched it to that.

PR #279, opened 2019-02-23
Array compare to handler

tsiq-karold:
Fixes https://github.com/twosigma/webtau/issues/151

### DIFF
--- a/webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/ArrayAndIterableCompareToHandler.java
+++ b/webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/ArrayAndIterableCompareToHandler.java
@@ -1,0 +1,92 @@
+package com.twosigma.webtau.expectation.equality.handlers;
+
+import com.twosigma.webtau.expectation.ActualPath;
+import com.twosigma.webtau.expectation.equality.CompareToComparator;
+import com.twosigma.webtau.expectation.equality.CompareToHandler;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ArrayAndIterableCompareToHandler implements CompareToHandler {
+    @Override
+    public boolean handleEquality(Object actual, Object expected) {
+        return actual.getClass().isArray() && expected instanceof Iterable;
+    }
+
+    @Override
+    public void compareEqualOnly(CompareToComparator comparator, ActualPath actualPath, Object actual, Object expected) {
+        List<?> actualList = toList(actual);
+
+        comparator.compareUsingEqualOnly(actualPath, actualList, expected);
+    }
+
+    private List<?> toList(Object actual) {
+        Class<?> componentType = actual.getClass().getComponentType();
+        if (componentType.equals(boolean.class)) {
+            return toList((boolean[]) actual);
+        } else if (componentType.equals(byte.class)) {
+            return toList((byte[]) actual);
+        } else if (componentType.equals(char.class)) {
+            return toList((char[]) actual);
+        } else if (componentType.equals(short.class)) {
+            return toList((short[]) actual);
+        } else if (componentType.equals(int.class)) {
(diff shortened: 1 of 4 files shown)


webtau-core/src/test/groovy/com/twosigma/webtau/expectation/equality/handlers/HandlerMessagesTest.groovy
```diff
@@ -232,6 +232,133 @@ class HandlerMessagesTest {
         } should throwException(AssertionError, ~/expected: not 1/)
     }
 
+    // ArrayAndIterableCompareToHandler messages
```
MykolaGolubyev: I can't remember the purpose of this super test class now :( Do we need to expand it with every new handler? Should we move code-throw cases to the handler test itself?

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/ArrayAndIterableCompareToHandler.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Override
+    public void compareEqualOnly(CompareToComparator comparator, ActualPath actualPath, Object actual, Object expected) {
+        List<?> actualList = toList(actual);
+
+        comparator.compareUsingEqualOnly(actualPath, actualList, expected);
+    }
+
+    private List<?> toList(Object actual) {
```
MykolaGolubyev: wow, this is pretty elaborate!

tsiq-karold: It's only because, as far as I'm aware, you really can't do anything terribly generic with primitive arrays.

webtau-core/src/test/groovy/com/twosigma/webtau/expectation/equality/handlers/HandlerMessagesTest.groovy
tsiq-karold: I don't remember the purpose either but as it seemed to have some tests for virtually all handlers I figured I should add some too.

I agree it would make more sense for these to live with the relevant handler tests but perhaps that can be a cleanup effort for a separate PR.

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/ArrayAndIterableCompareToHandler.java
MykolaGolubyev: @tsiq-karold what about Arrays.steam.box or something like that?

webtau-core/src/test/groovy/com/twosigma/webtau/expectation/equality/handlers/HandlerMessagesTest.groovy
MykolaGolubyev: I feel like it had handlers, to make sure they are actually wired, but I started to do it in each of the handlers themselves. Feel free to move the one you added to its own test.

webtau-core/src/main/java/com/twosigma/webtau/expectation/equality/handlers/ArrayAndIterableCompareToHandler.java
tsiq-karold: I use that where I can but `Array.stream()` only has overrides for doubles, ints and longs.

tsiq-nick: Ah, just noticed. Weird.

tsiq-karold: Yes, it is: https://stackoverflow.com/questions/754294/convert-an-array-of-primitive-longs-into-a-list-of-longs#comment35809264_23117638

Even if there were other overrides, it would still require the switch statement and cast.  Would just mean I didn't have to write specific `toList` functions.

PR #280, opened 2019-02-23
NullCompareTo handler deals with greater-less. Modify mismatch messages.

### LINKED ISSUE #259: comparing `null > number` gives confusing error
There is no compareTo handler that deals with numbers and `null`. Mismatch becomes an Error instead of assertion failure.

MykolaGolubyev:
This closes #259.

Since we are more precise in mismatch reporting now, Equal and Greater/Less Matchers do not provide summary text in case of mismatch anymore. Instead they solely rely on the generated Comparator mismatch report. 
Main motivation for the change is comparisons like `(null|value).shouldBe > (number|null)`. Matcher would provide additional message like "null is less than number" which would be inaccurate in case of nulls.


MykolaGolubyev:
@tsiq-karold actually the other way around. Because you added those features I could remove the summary text.

PR #285, opened 2019-03-05
junit5 dynamic tests from TableData

MykolaGolubyev:
```
class DynamicTestsGroovyTest {
    @TestFactory
    def priceCalculator() {
        ["price", "quantity", "outcome"] {
        _________________________________
          10     |  30      |  300
          -10    |  30      | -300
        }.useCases {
            PriceCalculator.calculate(price, quantity).should == outcome
        }
    }
}
```

### DIFF
--- a/webtau-junit5/src/main/java/com/twosigma/webtau/junit5/DynamicTests.java
+++ b/webtau-junit5/src/main/java/com/twosigma/webtau/junit5/DynamicTests.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2019 TWO SIGMA OPEN SOURCE, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.twosigma.webtau.junit5;
+
+import com.twosigma.webtau.data.table.Record;
+import com.twosigma.webtau.data.table.TableData;
+import org.junit.jupiter.api.DynamicTest;
+
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+
+public class DynamicTests {
+    private DynamicTests() {
+    }
+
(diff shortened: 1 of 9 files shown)


webtau-junit5/src/main/java/com/twosigma/webtau/junit5/DynamicTests.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+
+public class DynamicTests {
+    private DynamicTests() {
+    }
+
+    public static Stream<DynamicTest> fromTable(String useCasesPrefix, TableData useCases, Consumer<Record> test) {
+        String displayNamePrefix = useCasesPrefix.isEmpty() ? "" : useCasesPrefix + ": ";
```
tsiq-clemens: should it be empty or the row number? I think the latter is the DynamicTest default for java.

tsiq-clemens: or maybe I misunderstand the display name approach (I'll take it for a spin). How about optionally designating a column as display name? If you have more than 2 columns or lengthier values, the names can easily get too long for meaningful display.

MykolaGolubyev:
@tsiq-karold, @tsiq-clemens having doubts about `useCases` function name. Should it be `forEachUseCase` or something like that?

webtau-junit5/src/main/java/com/twosigma/webtau/junit5/DynamicTests.java
MykolaGolubyev: display name for each case will be `Record.toString()`. Prefix is optional text to put in front. 
I thought about `display` like column but then ran tests from IDEA and it was clear to me what is going on. Also if you want a row to have a dedicated test name, maybe it should be its own test method?

tsiq-karold:
`forEachTestCase`?

MykolaGolubyev:
The only confusion part I see with `forEach...` is that `forEach` semantic is not to return anything. Current `useCases` actually returns a `Stream<DynamicTest>`. 
Should it then be `mapTestCase`?

tsiq-karold:
Hmm, that's fair.  How about `testCode`?

MykolaGolubyev:
`test`?

tsiq-karold:
I'm happy with that.

webtau-junit5/src/main/java/com/twosigma/webtau/junit5/DynamicTests.java
tsiq-clemens: As discussed, we'll add an optional "label" column feature in a separate PR

MykolaGolubyev:
`.validate` ?

tsiq-clemens:
or `.check` ?

MykolaGolubyev:
OK, settling down on `test`. Maybe we will add aliases in future.

PR #286, opened 2019-03-06
fix greater/less matchers toString impl

### LINKED ISSUE #278: NullCompareTo handler doesn't play nice with expected being another matcher.
In case expected is another matcher, NullCompareTo handler should delegate to that matcher, but only if it knows how to handle nulls itself. 

```
def map = [a: null, b: null]
map.should == [a: greaterThan(100), b: lessThan(50)]
```

MykolaGolubyev:
closes #278


MykolaGolubyev:
It is subtle. NullCompareToHandler will not call matcher on the expected site. matcher will not be triggered and therefore comparator instance will be null. 

As I am explaining this, I wonder if nullcompareto handler should call matcher, but only if that matcher also handlers nulls.

tsiq-karold:
Got it.  The only new test I see is essentially:
```
    @Test
    void "rendering as string should render matching logic and expected value"() {
        assert matcher.toString() == '<greater than 8>'
    }
```
That does not look like it would trigger null comparators, would it?  Do we need another test?

MykolaGolubyev:
@tsiq-karold NullComparator was a case where a wrong toString impl emerged. Will add a test as a regression check.

PR #291, opened 2019-03-27
add table data permutation for input prep

MykolaGolubyev:
```
["Col A"              | "Col B"         | "Col C"] {
___________________________________________________________
 permute(true, false) | "v1b"           | permute('a', 'b')
 "v2a"                | permute(10, 20) | "v2c" }
```


webtau-core/src/test/groovy/com/twosigma/webtau/data/table/TableDataTest.groovy
```diff
@@ -38,4 +38,19 @@ class TableDataTest {
                  "v1a",   "v1b", "v1c",
                  "v2a",   "v2b")
     }
+
+    @Test
+    void "should generate multiple raws from multi-values"() {
```
tsiq-clemens: s/raws/rows

PR #296, opened 2019-04-01
Helper for query params construction

tsiq-karold:


webtau-docs/webtau/REST/query-parameters.md
```diff
@@ -0,0 +1,17 @@
...
+
+:include-groovy: com/twosigma/webtau/http/HttpGroovyTest.groovy {entry: "query params in url", bodyOnly: true}
+
+# Map based `query` helper
+
+This version of the `http.query` helper is generally more suitable for Groovy.
+
+:include-groovy: com/twosigma/webtau/http/HttpGroovyTest.groovy {entry: "build query params using map helper", bodyOnly: true}
+
+# Vararg based `query` helper
```
tsiq-clemens: s/Vararg/Varargs

webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
```diff
@@ -174,6 +174,14 @@ class HttpGroovyTest implements HttpConfiguration {
         }
     }
 
+    @Test
+    void "query params in url"() {
+        http.get("params?a=1&b=text") {
```
tsiq-clemens: just to be mean: what happens if you repeat the "a" key, e.g. `params?a=1&a=x`

webtau-http/src/test/java/com/twosigma/webtau/http/HttpJavaTest.java
```diff
@@ -238,6 +239,25 @@ public void redirectPut() {
         });
     }
 
+    @Test
```
tsiq-clemens: is there anything queryParam-related property we can test for to validate it honored the parameter as expected?

webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
```diff
@@ -183,6 +191,36 @@ class HttpGroovyTest implements HttpConfiguration {
...
+        http.get("params", http.query([a: 1, b: 'text'])) {
+            a.should == 1
+            b.should == 'text'
+        }
+    }
+
+    @Test
+    void "build query params using var arg helper"() {
+        http.get("params", http.query('a', '1', 'b', 'text')) {
+            a.should == 1
```
MykolaGolubyev: this example may be confusing for docs usage, as it is not clear why `a` should equal 1. 
Headers documentation just use `// assertion goes here`. That means that you may need another test to just validate syntax, but have no actual validation.

webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
tsiq-karold: The test service used by these tests works on matching url and the one that's been setup for this is `/params?a=1&b=text`.  I could add one with repeated params but that's not a realistic test of a "real" service.

That said, I'm pretty sure the http code in webtau (and in HttpUrlConnection) will handle that correctly if listed in the url string.  I don't think webtau's `HttpQueryParams` class can handle that though.  If we feel it should, I suggest we tackle it in a separate PR.

webtau-http/src/test/java/com/twosigma/webtau/http/HttpJavaTest.java
tsiq-karold: I'll change this to the params version because the test service is set up to respond with success to the params resource only if the appropriate query params are specified.

webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
tsiq-karold: Updated

tsiq-karold:
Can I get another review please? @MykolaGolubyev @tsiq-clemens

PR #298, opened 2019-04-03
introduce shortcut for cell.above

MykolaGolubyev:
share test validation between java and groovy
hide cell.above.plus from docs as it is not extensible yet

```
["Name" | "Start Date"              | "Games To Play" ] {
 ______________________________________________________
 "John" | LocalDate.of(2016, 6, 20) | 10
 "Bob"  | _________________         |  8
 "Mike" | _________________         | 14

 "Drew" | LocalDate.of(2016, 6, 22) | 10
 "Pete" | _________________         | 11
 "Max"  | _________________         |  3 }
```

**UPDATE**: removed `___` as a shortcut. Leaving this PR for merge and review as it has an important test org refactor.


webtau-docs/webtau/reference/table-data.md
```diff
@@ -29,24 +29,36 @@ Use `permute(v1, v2)` to automatically generate multiple rows.
...
\ No newline at end of file
+:include-table: table-with-cell-above.json
+
+You can use underscores of various length as a shortcut for `cell.above`
+
+```tabs
+Groovy:
+:include-groovy: com/twosigma/webtau/data/table/TableDataGroovyTest.groovy {entry: "createTableDataWithAboveRefShortcut", bodyOnly: true}
+
+ Note: Unlike Java, Groovy can use underscores of any length as long as it is longer than 1 character.
```
tsiq-clemens: I'd suggest a slight rewording:
In Groovy, unlike Java, you can use underscores of any length for the shortcut as long as ...

MykolaGolubyev: Thank you

webtau-core/src/main/java/com/twosigma/webtau/Ddjt.java
```diff
@@ -46,18 +47,18 @@
...
-    public static final TableDataUnderscoreOrPlaceholder ________________________________ = TableDataUnderscoreOrPlaceholder.INSTANCE;
-    public static final TableDataUnderscoreOrPlaceholder ________________________________________ = TableDataUnderscoreOrPlaceholder.INSTANCE;
-    public static final TableDataUnderscoreOrPlaceholder ________________________________________________ = TableDataUnderscoreOrPlaceholder.INSTANCE;
-    public static final TableDataUnderscoreOrPlaceholder ________________________________________________________________ = TableDataUnderscoreOrPlaceholder.INSTANCE;
-    public static final TableDataUnderscoreOrPlaceholder ________________________________________________________________________________ = TableDataUnderscoreOrPlaceholder.INSTANCE;
-    public static final TableDataUnderscoreOrPlaceholder ________________________________________________________________________________________________ = TableDataUnderscoreOrPlaceholder.INSTANCE;
-
     public static final TableDataCellValueGenFunctions cell = new TableDataCellValueGenFunctions();
 
+    public static final TableDataCellValueGenerator<?> __ = TableDataCellAbove.generator;
```
tsiq-karold: Not sure how I feel about the _____'s.  When I see that in the table, I don't immediately think that means value from above.  To me they indicate absence of value.

In written form, typically you'd do `-"-` but obviously that's not possible here so while I don't really like this I'm not sure I can think of a better alternative 😞 

Having said that, do we really need a shortcut?

MykolaGolubyev: @tsiq-karold to me it felt non ambiguous, but if you feel that way there is a pretty good chance that it is not clear for others either. 
The main rationale for `_____` is it sort of let you see through important values only and is easier to glance over than `cell.above`. 

@tsiq-clemens what is your first reaction on `____`?

Pulling @markroth8 for additional opinion.

tsiq-karold: Are we not using `_____` to also separate headers from data rows?  I vaguely recall a PR for that.

MykolaGolubyev: There is still code to handle this.

tsiq-karold: I understand that, I was referring to the fact that the underscores now mean two different things though I guess the context will generally make it clear which it is.

MykolaGolubyev: Got a piece of a feedback. Immediate reaction is "value is empty", but with documentation it makes sense.

PR #301, opened 2019-04-12
move TestFactory doc to junit 5 runner 

tsiq-clemens:
it is not groovy-runner specific as @MykolaGolubyev pointed out


webtau-docs/webtau/generic-runners/JUnit-5.md
```diff
@@ -11,3 +11,16 @@ Java: :include-file: com/example/tests/junit5/CustomerCrudSeparatedJavaTest.java
 # Maven Import
 
 :include-file: maven/junit5-dep.xml
+
+# Specific Features
+## TestFactory
```
MykolaGolubyev: Lets make make TestFactory first level title.

tsiq-clemens: done

PR #303, opened 2019-04-15
Undeclared throwable

tsiq-clemens:
Some code flows involve a proxy that wrap checked exceptions in an UndeclaredThrowableException (see https://docs.oracle.com/javase/8/docs/api/index.html?java/lang/reflect/UndeclaredThrowableException.html) .
In this case, we still need to extract the wrapped exception class and message.

### DIFF
--- a/webtau-core/src/main/java/com/twosigma/webtau/expectation/code/ThrowExceptionMatcher.java
+++ b/webtau-core/src/main/java/com/twosigma/webtau/expectation/code/ThrowExceptionMatcher.java
@@ -16,13 +16,13 @@
 
 package com.twosigma.webtau.expectation.code;
 
+import static com.twosigma.webtau.Ddjt.createActualPath;
+
 import com.twosigma.webtau.expectation.CodeBlock;
 import com.twosigma.webtau.expectation.CodeMatcher;
 import com.twosigma.webtau.expectation.equality.CompareToComparator;
-
+import java.lang.reflect.UndeclaredThrowableException;
 import java.util.regex.Pattern;
-
-import static com.twosigma.webtau.Ddjt.createActualPath;
 
 public class ThrowExceptionMatcher implements CodeMatcher {
     private String expectedMessage;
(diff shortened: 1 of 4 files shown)


webtau-core/src/main/java/com/twosigma/webtau/expectation/code/ThrowExceptionMatcher.java
```diff
@@ -98,4 +97,15 @@ public boolean matches(CodeBlock codeBlock) {
 
         return isEqual;
     }
+
+    private void extractExceptionDetails(Throwable t) {
+        if (t instanceof UndeclaredThrowableException) {
+            Throwable undeclaredCheckedException = t.getCause();
```
tsiq-karold: Could cause ever be null?

webtau-core/src/test/groovy/com/twosigma/webtau/expectation/code/ThrowExceptionMatcherGroovyTest.groovy
```diff
@@ -119,9 +121,20 @@ class ThrowExceptionMatcherGroovyTest {
         } should throwException(IllegalArgumentException)
     }
 
+    @Test
+    void "handle undeclared throwable exceptions"() {
+        code { checkedExceptionThrowingProxy('catch me if you can') }
```
tsiq-karold: 😁

webtau-core/src/main/java/com/twosigma/webtau/expectation/code/ThrowExceptionMatcher.java
tsiq-clemens: Not if the proxy does its job correctly, I think. The docs for the exception say 

> As of release 1.4, this exception has been retrofitted to conform to the general purpose exception-chaining mechanism. The "undeclared checked exception that was thrown by the invocation handler" that may be provided at construction time and accessed via the getUndeclaredThrowable() method is now known as the cause, and may be accessed via the Throwable.getCause() method, as well as the aforementioned "legacy method."

PR #305, opened 2019-04-15
capture file parse failures as "init" failed test

### LINKED ISSUE #289: Exceptions outside of scenario should be caught, not cause webtau to fail
# Steps

```
package scenarios

import static com.twosigma.webtau.WebTauGroovyDsl.scenario

throw new RuntimeException("kaboom")

scenario("failure") {
    true.should == false
}
```

When this scenario is run, no output report is given, remaining tests aren't run (if any).

Expected behavior: maybe entire file should be marked as a failure, no tests run. or something else to that effect and is reasonable.

MykolaGolubyev:
closes #289

### DIFF
--- a/webtau-standalone-runner-groovy/src/main/groovy/com/twosigma/webtau/runner/standalone/StandaloneTestRunner.groovy
+++ b/webtau-standalone-runner-groovy/src/main/groovy/com/twosigma/webtau/runner/standalone/StandaloneTestRunner.groovy
@@ -32,8 +32,9 @@ class StandaloneTestRunner {
 
     private AtomicBoolean isTerminated
 
-    private ThreadLocal<TestRunCondition> runCondition = ThreadLocal.<TestRunCondition>withInitial { ->
-        new TestRunCondition(isConditionMet: true) }
+    private ThreadLocal<TestRunCondition> runCondition = ThreadLocal.<TestRunCondition> withInitial { ->
+        new TestRunCondition(isConditionMet: true)
+    }
 
     StandaloneTestRunner(GroovyScriptEngine groovy, Path workingDir) {
         this.workingDir = workingDir.toAbsolutePath()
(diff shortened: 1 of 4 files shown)


MykolaGolubyev:
@tsiq-karold I think you got it. Scenario is only added in case of parse failure. Need to test http.calls part, but I would assume so.

webtau-standalone-runner-groovy/src/main/groovy/com/twosigma/webtau/runner/standalone/StandaloneTestRunner.groovy
```diff
@@ -47,16 +48,24 @@ class StandaloneTestRunner {
...
+            script.setProperty("sscenario", this.&sscenario)
+            script.setProperty("onlyWhen", this.&onlyWhen)
 
-        StandaloneTestListeners.beforeScriptParse(scriptPath)
-        script.run()
+            script.run()
+        })
+
+        scriptParse.run()
+        if (scriptParse.hasError()) {
```
MykolaGolubyev: May need to add check for assertion if anyone decides to add assertion outside of scenario.

tsiq-karold:
Is it worth showing the scenario if there are no failures but there are http calls?  It may make more sense to have those show up under the init than within the first scenario.

MykolaGolubyev:
Frankly, I am considering throwing unsupported operation exception if someone tries to call http outside of scenario. Things outside of scenario will be out of order during parallel execution. And they will run across all the files first and then scenarios will kick-in.

tsiq-karold:
Actually, that's a fair point.  What I actually had in mind is lazy resources but those will execute in the context of a scenario.  I think it makes sense to disallow http outside a scenario.

MykolaGolubyev:
Added #306. Should we merge this part for now?

PR #315, opened 2019-04-27
add dynamic math. add table data cell.above plus/minus doc

### LINKED ISSUE #299: Extensible rules for `add`, `subtract`, `plus`, `minus`
With introduction of `TableData`'s `cell.above.plus(x)` we need a way to extend the math for various types in a way that it works with Java. E.g. `_____.plus(1)` or `___ + 1` (for Groovy) where `____` defined as `LocalDate.of(...)`.

We also need to expose this math outside of TableData cells. So something like `add(localDate, 1)` should be available. 

Rules need to be discoverable in a similar manner how `CompareToHandler`s are. A handful of rules should be available out of the box.

MykolaGolubyev:
closes #299

### DIFF
--- a/webtau-core/src/test/groovy/com/twosigma/webtau/data/math/DynamicMathTest.groovy
+++ b/webtau-core/src/test/groovy/com/twosigma/webtau/data/math/DynamicMathTest.groovy
@@ -1,0 +1,17 @@
+package com.twosigma.webtau.data.math
+
+import org.junit.Test
+
+class DynamicMathTest {
+    @Test
+    void "should use discovered simple math handlers for add operation"() {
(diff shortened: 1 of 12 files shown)


webtau-core/src/test/groovy/com/twosigma/webtau/data/math/DynamicMathTest.groovy
```diff
@@ -0,0 +1,17 @@
+package com.twosigma.webtau.data.math
+
+import org.junit.Test
+
+class DynamicMathTest {
```
tsiq-clemens: want to add tests for incompatible operands?

PR #319, opened 2019-04-30
Recursive scenario file discovery

### LINKED ISSUE #318: Recursively search for scenario files
For large test suites, having a single scenarios directory is not practical.  One can specify multiple patterns on the CLI and in Maven but if one does that in Maven we lose the ability to override it with a -D at run time.

tsiq-karold:
Closes #318


MykolaGolubyev:
oh nice! maybe use kebab-case-for-test-dirs to match file-names?

tsiq-karold:
Hmm, actually I should change the file names because my examples wouldn't actually work.  IIRC, the file names need to match allowed chars for a java class.

PR #322, opened 2019-05-07
consolidate ServiceUtils and ServiceLoaderUtils

MykolaGolubyev:


MykolaGolubyev:
@tsiq-karold was trying to remember myself why one version was set. My best guess is helped to prevent duplicates.

PR #333, opened 2019-05-17
Send user agent

### LINKED ISSUE #329: Set a webtau user agent by default and allow overriding in cfg

tsiq-karold:
Closes #329 

Default behaviour is to send `webtau/<version>`.  If a user agent is specified in cfg, we send `<cfg-user-agent> (webtau/<version>)`.  It is also possible to not append the `(webtau/<version>)` bit with a custom agent by specifying an appropriate cfg option.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -610,6 +610,19 @@
                     </dependency>
                 </dependencies>
             </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifest>
+                            <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
+                            <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
+                        </manifest>
+                    </archive>
+                </configuration>
+            </plugin>
         </plugins>
     </build>
 
--- a/webtau-config/src/main/java/com/twosigma/webtau/cfg/WebTauConfig.java
+++ b/webtau-config/src/main/java/com/twosigma/webtau/cfg/WebTauConfig.java
@@ -47,7 +47,10 @@ public class WebTauConfig {
     private final ConfigValue disableFollowingRedirects = declareBoolean("disableRedirects", "disable following of redirects from HTTP calls");
     private final ConfigValue maxRedirects = declare("maxRedirects", "Maximum number of redirects to follow for an HTTP call", () -> 20);
     private final ConfigValue userAgent = declare("userAgent", "User agent to send on HTTP requests",
-            () -> "webtau/"+getClass().getPackage().getImplementationVersion());
+            () -> "webtau/" + getClass().getPackage().getImplementationVersion());
+    private final ConfigValue removeWebtauFromUserAgent = declare("removeWebtauFromUserAgent",
+            "By default webtau appends webtau and its version to the user-agent, this disables that part",
+            () -> false);
     private final ConfigValue workingDir = declare("workingDir", "logical working dir", () -> Paths.get(""));
     private final ConfigValue cachePath = declare("cachePath", "user driven cache file path",
             () -> workingDir.getAsPath().resolve(".webtau.cache.json"));
@@ -187,12 +190,21 @@ public class WebTauConfig {
             return userAgent.getAsString();
         }
 
-        String defaultValue = userAgent.getDefaultValue().toString();
-        return userAgent.getAsString() + "/" + defaultValue;
+        String finalUserAgent = userAgent.getAsString();
+        if (!removeWebtauFromUserAgent.getAsBoolean()) {
+            String defaultValue = userAgent.getDefaultValue().toString();
+            finalUserAgent += " (" + defaultValue + ")";
+        }
+
+        return finalUserAgent;
     }
 
     public ConfigValue getUserAgentConfigValue() {
         return userAgent;
+    }
+
+    public ConfigValue getRemoveWebtauFromUserAgent() {
+        return removeWebtauFromUserAgent;
     }
 
     public Path getDocArtifactsPath() {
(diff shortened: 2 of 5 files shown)


tsiq-rob:
> I thought it would be useful to include the webtau part and version even with a custom agent.

as a counter argument, this will probably annoy someone who specifically wants to override the user agent with something of their own choosing

tsiq-karold:
Yeah, that's fair.  I considered allowing both but I don't want the cfg to get too crazy.

webtau-config/src/main/java/com/twosigma/webtau/cfg/WebTauConfig.java
```diff
@@ -46,6 +46,8 @@
     private final ConfigValue waitTimeout = declare("waitTimeout", "wait timeout in milliseconds", () -> 5000);
     private final ConfigValue disableFollowingRedirects = declareBoolean("disableRedirects", "disable following of redirects from HTTP calls");
     private final ConfigValue maxRedirects = declare("maxRedirects", "Maximum number of redirects to follow for an HTTP call", () -> 20);
+    private final ConfigValue userAgent = declare("userAgent", "User agent to send on HTTP requests",
+            () -> "webtau/"+getClass().getPackage().getImplementationVersion());
```
tsiq-clemens: a couple of spaces around the + make it prettier

webtau-config/src/main/java/com/twosigma/webtau/cfg/WebTauConfig.java
```diff
@@ -180,6 +182,19 @@ public int maxRedirects() {
         return maxRedirects.getAsInt();
     }
 
+    public String getUserAgent() {
+        if (userAgent.isDefault()) {
+            return userAgent.getAsString();
+        }
+
+        String defaultValue = userAgent.getDefaultValue().toString();
+        return userAgent.getAsString() + "/" + defaultValue;
```
tsiq-clemens: would it look better to have the webtau/version in (), e.g.   partner-api-tests (webtau/1.8).
vs partner-api-tests/webtau/1.8, or what if people have a slash in their agent config? api/v2/webtau/1.8

I might be overthinking this.

tsiq-karold: I think you might be right, I prefer it in () than after a /.

tsiq-karold:
Hopefully this latest version satisfies both @tsiq-clemens and @tsiq-rob 's requirements/comments 😄

pom.xml
```diff
@@ -610,6 +610,19 @@
                     </dependency>
                 </dependencies>
             </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
```
MykolaGolubyev: what is this for?

webtau-config/src/main/java/com/twosigma/webtau/cfg/WebTauConfig.java
```diff
@@ -46,6 +46,11 @@
     private final ConfigValue waitTimeout = declare("waitTimeout", "wait timeout in milliseconds", () -> 5000);
     private final ConfigValue disableFollowingRedirects = declareBoolean("disableRedirects", "disable following of redirects from HTTP calls");
     private final ConfigValue maxRedirects = declare("maxRedirects", "Maximum number of redirects to follow for an HTTP call", () -> 20);
+    private final ConfigValue userAgent = declare("userAgent", "User agent to send on HTTP requests",
+            () -> "webtau/" + getClass().getPackage().getImplementationVersion());
```
MykolaGolubyev: I think we expose version in some other way

webtau-config/src/main/java/com/twosigma/webtau/cfg/WebTauConfig.java
```diff
@@ -46,6 +46,11 @@
...
+    private final ConfigValue removeWebtauFromUserAgent = declare("removeWebtauFromUserAgent",
```
MykolaGolubyev: was wondering why enable is a toggle and disable is default. It is a bit unusual for REST api to expect user agent.

webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
```diff
@@ -239,6 +240,41 @@ class HttpGroovyTest implements HttpConfiguration {
...
+        } finally {
+            cfg.userAgentConfigValue.reset()
+        }
+    }
+
+    @Test
+    void "custom user agent without webtau and its version"() {
+        try {
+            cfg.userAgentConfigValue.set("test", "custom")
+            cfg.removeWebtauFromUserAgent.set("true", true)
```
MykolaGolubyev: first parameter value is the source, not sure what "true" means in this case

MykolaGolubyev:
A bit late to the party, but left a few comments.

pom.xml
tsiq-karold: To have the version populated in the jar manifest and available via `getClass().getPackage().getImplementationVersion())`

webtau-config/src/main/java/com/twosigma/webtau/cfg/WebTauConfig.java
tsiq-karold: I didn't find it but if it is already available I can undo the jar packaging change and switch this line over.

webtau-config/src/main/java/com/twosigma/webtau/cfg/WebTauConfig.java
tsiq-karold: I agree that it's unlikely an API will expect an agent but it's useful for identifying requests from tests in logs.

webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
tsiq-karold: Oops, meant to say `test` like the line above.

PR #338, opened 2019-05-20
Report status code before schema

### LINKED ISSUE #302: Assertion failure reporting precedence
If a particular request fails both the status code assertion and Swagger spec validation, the report shows spec validation failure first.  I think by and large showing the status code assertion failure first would be more useful when debugging.

tsiq-karold:
Closes #302 

Before merging this, I'd like some feedback.  Prior to this PR, any `HttpValidationHandler` errors are reported even if status code fails or user expectations fail.  This is confusing in the case where you expect a request to succeed but it fails and the failure response is of a different schema to what you'd expect; it's confusing because you get a schema validation failure and the fact the request failed is not reported.

We have two options.  The behaviour this PR introduces is to only run `HttpValidationHandlers` if everything else is fine.  So generally you'll get a status code error or user expectation error and only if both of those are fine will you get a schema error.

Do you think it would be better to report in this order:
1. status code mismatch
1. schema failures
1. user expectations

I suspect that's generally better because if the schema is incorrect the odds of the user expectations passing are reduced anyway and could send someone down a rabbit hole.  Unfortunately, this probably requires substantially more refactoring.

Thoughts?

### DIFF
--- a/webtau-feature-testing/test-expectations/scenarios/rest/openapi/disableOpenApiValidation/run-details.json
+++ b/webtau-feature-testing/test-expectations/scenarios/rest/openapi/disableOpenApiValidation/run-details.json
@@ -7,7 +7,7 @@
   }, {
     "scenario" : "disable request validation",
     "stepsSummary" : {
-      "numberOfFailed" : 2
+      "numberOfSuccessful" : 3
     }
   }, {
     "scenario" : "disable response validation",
@@ -15,5 +15,5 @@
       "numberOfSuccessful" : 3
     }
   } ],
-  "exitCode" : 1
+  "exitCode" : 0
 }
--- a/webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
+++ b/webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
@@ -23,6 +23,7 @@ import com.twosigma.webtau.http.datanode.DataNode
 import com.twosigma.webtau.http.datanode.GroovyDataNode
 import com.twosigma.webtau.http.testserver.TestServerResponse
 import com.twosigma.webtau.http.validation.HttpResponseValidator
+import com.twosigma.webtau.http.validation.HttpValidationHandlers
 import com.twosigma.webtau.utils.ResourceUtils
 import com.twosigma.webtau.utils.UrlUtils
 import org.junit.*
@@ -980,6 +981,18 @@ class HttpGroovyTest implements HttpConfiguration {
         } should throwException(HttpException, ~/error during http\.get/)
 
         http.lastValidationResult.errorMessage.should == ~/java.lang.IllegalArgumentException: Request header is null/
+    }
+
+    @Test
+    void "status code mismatch reported before additional validators"() {
+        HttpValidationHandlers.register { result -> throw new AssertionError((Object)"schema validation error") }
+        try {
+            code {
+                http.get("/notfound") {}
+            } should throwException(AssertionError, ~/(?s)header.statusCode.*404.*200/)
+        } finally {
+            HttpValidationHandlers.reset()
+        }
     }
 
     @Override
--- a/webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
+++ b/webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
@@ -984,11 +984,39 @@ class HttpGroovyTest implements HttpConfiguration {
     }
 
     @Test
-    void "status code mismatch reported before additional validators"() {
+    void "implicit status code mismatch reported before additional validators"() {
         HttpValidationHandlers.register { result -> throw new AssertionError((Object)"schema validation error") }
         try {
             code {
                 http.get("/notfound") {}
+            } should throwException(AssertionError, ~/(?s)header.statusCode.*404.*200/)
+        } finally {
+            HttpValidationHandlers.reset()
+        }
+    }
+
+    @Test
+    void "explicit status code mismatch reported before additional validators"() {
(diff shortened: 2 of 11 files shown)


webtau-feature-testing/test-expectations/scenarios/rest/openapi/disableOpenApiValidation/run-details.json
```diff
@@ -7,13 +7,13 @@
   }, {
     "scenario" : "disable request validation",
     "stepsSummary" : {
-      "numberOfFailed" : 2
+      "numberOfSuccessful" : 3
```
MykolaGolubyev: do we still need a failed case?

webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
```diff
@@ -982,6 +983,75 @@ class HttpGroovyTest implements HttpConfiguration {
         http.lastValidationResult.errorMessage.should == ~/java.lang.IllegalArgumentException: Request header is null/
     }
 
+    @Test
+    void "implicit status code mismatch reported before additional validators"() {
+        HttpValidationHandlers.register { result -> throw new AssertionError((Object)"schema validation error") }
```
MykolaGolubyev: wonder if we should use the `withAdditional...` pattern we use for some other handler like classes

webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
```diff
@@ -982,6 +983,75 @@ class HttpGroovyTest implements HttpConfiguration {
...
+                    id.should == 'foo'
+                }
+            } should throwException(AssertionError, ~/(?s)header.statusCode.*404.*200/)
+        } finally {
+            HttpValidationHandlers.reset()
+        }
+    }
+
+    @Test
+    void "assertion and additional validation errors"() {
```
MykolaGolubyev: style of the test name mismatches the other ones.

webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
```diff
@@ -982,6 +983,75 @@ class HttpGroovyTest implements HttpConfiguration {
...
+                }
+            } should throwException(AssertionError, ~/expected: "foo"/)
+        } finally {
+            HttpValidationHandlers.reset()
+        }
+    }
+
+    @Test
+    void "additional validator errors reported if status code is correct"() {
+        HttpValidationHandlers.register { result -> throw new AssertionError((Object)"schema validation error") }
```
MykolaGolubyev: if (as per other comment) we don't introduce a `withAdditional` like method, then maybe we should extract the repetitive code in this test to a private static method `wtihAdditionalFailedHandler('schema validation error') {...}`

webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
```diff
@@ -982,6 +983,75 @@ class HttpGroovyTest implements HttpConfiguration {
...
+    @Test
+    void "assertion and additional validation errors"() {
+        HttpValidationHandlers.register { result -> throw new AssertionError((Object)"schema validation error") }
+        try {
+            code {
+                http.get("/notfound") {
+                    statusCode.should == 404
+                    id.should == 'foo'
+                }
+            } should throwException(AssertionError, ~/expected: "foo"/)
```
MykolaGolubyev: not clear by this check if the `schema validation error` is not part of the exception.

webtau-http/src/main/java/com/twosigma/webtau/http/validation/HttpValidationHandlers.java
```diff
@@ -23,6 +23,15 @@
 public class HttpValidationHandlers {
     private static final List<HttpValidationHandler> configurations = ServiceLoaderUtils.load(HttpValidationHandler.class);
 
+    public static void register(HttpValidationHandler handler) {
```
MykolaGolubyev: Most of the times I call these methods simply `add`. But I may have slipped a few ones like `register`. Should we rename to `add`?

webtau-http/src/main/java/com/twosigma/webtau/http/validation/HttpValidationHandlers.java
```diff
@@ -23,6 +23,15 @@
...
+        configurations.add(handler);
+    }
+
+    public static void reset() {
```
MykolaGolubyev: Would prefer to maintain as fewer `resets` like methods as possible.

webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
```diff
@@ -982,6 +983,75 @@ class HttpGroovyTest implements HttpConfiguration {
         http.lastValidationResult.errorMessage.should == ~/java.lang.IllegalArgumentException: Request header is null/
     }
 
+    @Test
+    void "implicit status code mismatch reported before additional validators"() {
+        HttpValidationHandlers.register { result -> throw new AssertionError((Object)"schema validation error") }
+        try {
+            code {
+                http.get("/notfound") {}
+            } should throwException(AssertionError, ~/(?s)header.statusCode.*404.*200/)
```
MykolaGolubyev: order is not clear based on the assertion

tsiq-karold:
I think I've addressed all your comments @MykolaGolubyev , please review again.

MykolaGolubyev:
beautiful! 
Do we still need to filter resources in pom.xml? If not, let's remove it to avoid confusion?

tsiq-karold:
I wondered about that and then forgot to check.  Looks safe to remove though, commit coming up.

PR #341, opened 2019-05-22
initial cli testing frame

MykolaGolubyev:
Initial DSL for cli testing. Not exposed through `WebTauDsl` yet.  `CliJavaTest` shows basic java syntax. Approach is similar to `Http.`. 

It is most likely going to change so considering it is a private API for now. Want to put it out there so we can start conversation around. 

One of the syntax I want to be able to achieve follows
```
cli.run('command...') {
    on('partial match') {
         http.get( ... ) { 
         }
    }

    onOptional('partial match') {
        http.del(....) {
        }
    }

    output.should contain('blah')
}
```

Alternatively we may try to use `waitTo` (not really sure how to approach this)
```
cli.run('command...') {
    line.waitTo == ~/partial match/ 
    http.get( ... ) { 
    }

    output.should contain('blah')
}
```

### DIFF
--- a/webtau-cli/src/main/java/com/twosigma/webtau/cli/Cli.java
+++ b/webtau-cli/src/main/java/com/twosigma/webtau/cli/Cli.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2019 TWO SIGMA OPEN SOURCE, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.twosigma.webtau.cli;
+
+import com.twosigma.webtau.cli.expectation.CliExitCode;
+import com.twosigma.webtau.cli.expectation.CliOutput;
+import com.twosigma.webtau.cli.expectation.CliValidationExitCodeOutputHandler;
+import com.twosigma.webtau.cli.expectation.CliValidationOutputOnlylHandler;
+import com.twosigma.webtau.reporter.TestStep;
+
+import java.util.function.Consumer;
+
+import static com.twosigma.webtau.reporter.IntegrationTestsMessageBuilder.action;
+import static com.twosigma.webtau.reporter.IntegrationTestsMessageBuilder.stringValue;
+import static com.twosigma.webtau.reporter.TokenizedMessage.tokenizedMessage;
+
+public class Cli {
+    public static final Cli cli = new Cli();
+
+    private Cli() {
+    }
+
+    public void run(String command, CliValidationOutputOnlylHandler handler) {
+        cliStep(command, (runResult) -> handler.handle(cliOutput(runResult), cliError(runResult)));
+    }
+
+    public void run(String command, CliValidationExitCodeOutputHandler handler) {
+        cliStep(command, (runResult) -> handler.handle(exitCode(runResult), cliOutput(runResult), cliError(runResult)));
+    }
+
+    private void cliStep(String command, Consumer<ProcessRunResult> validationCode) {
+        TestStep.createAndExecuteStep(null,
+                tokenizedMessage(action("running cli command "), stringValue(command)),
+                () -> tokenizedMessage(action("ran cli clommand"), stringValue(command)),
+                () -> {
+                    ProcessRunResult runResult = ProcessUtils.run(command.split("\\s+"));
+                    validationCode.accept(runResult);
+                });
+    }
+
+    private CliExitCode exitCode(ProcessRunResult runResult) {
+        return new CliExitCode(runResult.getExitCode());
+    }
+
+    private static CliOutput cliOutput(ProcessRunResult runResult) {
+        return new CliOutput("out", runResult.getOutput());
+    }
+
+    private static CliOutput cliError(ProcessRunResult runResult) {
+        return new CliOutput("error", runResult.getError());
+    }
+}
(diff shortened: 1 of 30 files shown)


tsiq-karold:
Can you explain what the desired behaviour of `on` and `onOptional` are?

webtau-cli/src/main/java/com/twosigma/webtau/cli/Cli.java
```diff
@@ -0,0 +1,66 @@
...
+    public void run(String command, CliValidationExitCodeOutputHandler handler) {
+        cliStep(command, (runResult) -> handler.handle(exitCode(runResult), cliOutput(runResult), cliError(runResult)));
+    }
+
+    private void cliStep(String command, Consumer<ProcessRunResult> validationCode) {
+        TestStep.createAndExecuteStep(null,
+                tokenizedMessage(action("running cli command "), stringValue(command)),
+                () -> tokenizedMessage(action("ran cli clommand"), stringValue(command)),
+                () -> {
+                    ProcessRunResult runResult = ProcessUtils.run(command.split("\\s+"));
```
tsiq-karold: Hmmm, what if someone wants to run `foo.sh "something that should be one arg and is therefore in quotes"`?

webtau-cli/src/main/java/com/twosigma/webtau/cli/ProcessUtils.java
```diff
@@ -0,0 +1,43 @@
...
+
+package com.twosigma.webtau.cli;
+
+import java.io.IOException;
+
+public class ProcessUtils {
+    private ProcessUtils() {
+    }
+
+    public static ProcessRunResult run(String[] command) {
```
tsiq-karold: I must have written almost identical code at least 10 times in my career :)

webtau-cli/src/main/java/com/twosigma/webtau/cli/ProcessUtils.java
```diff
@@ -0,0 +1,43 @@
...
+        try {
+            Process process = processBuilder.start();
+
+            StreamGobbler outputGobbler = new StreamGobbler("output", process.getInputStream());
+            StreamGobbler errorGobbler = new StreamGobbler("error", process.getErrorStream());
+
+            new Thread(errorGobbler).start();
+            new Thread(outputGobbler).start();
+
+            process.waitFor();
```
tsiq-karold: Do we want to impose some timeouts here?

webtau-cli/src/main/java/com/twosigma/webtau/cli/ProcessUtils.java
```diff
@@ -0,0 +1,43 @@
...
+
+            return new ProcessRunResult(process.exitValue(), outputGobbler.getLines(), errorGobbler.getLines());
```
tsiq-karold: Could the reason for your sporadic test failures be that you're not joining the two gobbler threads?  They could in theory still be consuming the stream when you call `getLines` here.

webtau-cli/src/main/java/com/twosigma/webtau/cli/Cli.java
MykolaGolubyev: good point, thanks

webtau-cli/src/main/java/com/twosigma/webtau/cli/ProcessUtils.java
MykolaGolubyev: I feel you :)

webtau-cli/src/main/java/com/twosigma/webtau/cli/ProcessUtils.java
MykolaGolubyev: Created issue earlier to track this. Want a skeleton down for now.

webtau-cli/src/main/java/com/twosigma/webtau/cli/ProcessUtils.java
MykolaGolubyev: Trying

MykolaGolubyev:
> Do we want to support user supplied environment variables for the process? It's not uncommon for scripts to expect some vars. People may even wish to set a PATH so they can run all the tests without FQN for the scripts they're executing.
> 
> Do we want to support multiple script executions? Right now if you do two cli.run calls, they'll have totally separate environments (or shells or whatever you want to call it).
> 
> My hunch is yes to the env vars and no to the multiple scripts but curious on your thoughts.

Yes for env vars. Not sure I follow multiple scripts part. Do you mean like semi colon separated or an explicit list of commands? I think if there is a good usecase I would take a look at it.

MykolaGolubyev:
> Can you explain what the desired behaviour of `on` and `onOptional` are?

`on` will fail if there was no match during entire process run. `onOptional` (better name is welcome) is not going to fail if there was no match, but still lets you react (e.g. send `http.delete` to clear state).

webtau-cli/src/main/java/com/twosigma/webtau/cli/ProcessUtils.java
MykolaGolubyev: You were right, thanks @tsiq-karold. Wonder if executor would have been better in this case.

tsiq-karold: That thought had occurred to me.  However, it won't prevent you from having to join conceptually, instead you'll call get on a future.  You'll also have to work out how to size the executor or I suppose let the JVM do that for you and hope it's never an issue (which for this usecase it likely won't be).

tsiq-karold:
> > Do we want to support user supplied environment variables for the process? It's not uncommon for scripts to expect some vars. People may even wish to set a PATH so they can run all the tests without FQN for the scripts they're executing.
> > Do we want to support multiple script executions? Right now if you do two cli.run calls, they'll have totally separate environments (or shells or whatever you want to call it).
> > My hunch is yes to the env vars and no to the multiple scripts but curious on your thoughts.
> 
> Yes for env vars. Not sure I follow multiple scripts part. Do you mean like semi colon separated or an explicit list of commands? I think if there is a good usecase I would take a look at it.

Not sure I can think of a use case right now but yes conceptually semi colon separated or && separated is what I was thinking.  Generally that'd b the same as executing multiple cli.run sequentially anyway because unless you source the script it'll run in a subshell and therefore won't touch your env.

tsiq-karold:
The env stuff looks good.  Did you want to tackle the args with spaces thing in this PR or create an issue and tackle later?  I'm happy to approve if we have an issue for it,

MykolaGolubyev:
Let's do a separate PR: https://github.com/twosigma/webtau/issues/346

Issue #221, opened 2018-12-07
Transformations for doc capture


tsiq-karold:
This should be used sparingly but for list responses for example, it's likely that in docs you'll only want one or two items.  Supporting transformations at capture time will allow one to use somewhat sophisticated logic for selecting what to show in docs.

tsiq-karold:
Discussed this with @MykolaGolubyev, we propose the syntax listed below.  This is based on an example response:
```json
{
   "ints": [1,2,3,4,5],
    "strings": ["a", "b", "ab", "abc", "ba"]
}
```

Simple response filter:
```groovy
http.doc.filterResponse('ints', anInt -> anInt > 3)
```
Would generate:
```json
{
   "ints": [4,5],
    "strings": ["a", "b", "ab", "abc", "ba"]
}
```

Multiple filters:
```groovy
http.doc
    .filterResponse('ints', anInt -> anInt > 3)
    .filterResponse('strings', str -> str.startsWith('a')
    .capture()
```
Would generate:
```json
{
    "ints": [4,5],
    "strings": ["a", "ab", "abc"]
}
```

The reason for the fluid API with `filterResponse` methods is to allow us to later add filtration of requests, headers, etc.

Such capture calls should capture the unfiltered body in the existing path/filename and the filtered body in a file with a `filtered` infix (i.e. `response.filtered.json`).

MykolaGolubyev:
Is this still relevant?

MykolaGolubyev:
@tsiq-karold ^

tsiq-karold:
It's relevant in that we still don't have support for it but I don't have a need for it any more.

PR #343, opened 2019-05-23
Path based access to datanode children

### LINKED ISSUE #282: Path based access to DataNode fields

tsiq-karold:
closes #282


tsiq-karold:
I haven't updated documentation yet.  I was initially thinking of putting it in https://twosigma.github.io/webtau/guide/REST/data-node but it's really Java specific.  I mean it would work in Groovy but it doesn't make sense to use it in Groovy.  The alternative is Junit pages.  WDYT?

MykolaGolubyev:
I think that even in Groovy case it can be useful if someone decides to get/build a path at runtime. Or have a path inside a config file (i.e. different versions of api to access temperature from a city).

webtau-http/src/main/java/com/twosigma/webtau/http/datanode/StructuredDataNode.java
```diff
@@ -59,19 +59,66 @@ public DataNodeId id() {
...
-                children.get(name):
+                children.get(name) :
                 new NullDataNode(id.child(name));
     }
 
+    private DataNode getIndexedChild(String name, int openBraceIdx, int closeBraceIdx) {
+        int additionalOpenIdx = name.indexOf('[', openBraceIdx + 1);
+        int additionalCloseId = name.indexOf(']', closeBraceIdx + 1);
+
+        if (additionalOpenIdx != -1 || additionalCloseId != -1 || openBraceIdx > closeBraceIdx) {
```
MykolaGolubyev: what about cases like `[[1]`

webtau-http/src/main/java/com/twosigma/webtau/http/datanode/StructuredDataNode.java
```diff
@@ -59,19 +59,66 @@ public DataNodeId id() {
...
+            throw new IllegalArgumentException("Requested name " + name + " contains mismatched indexing brackets");
+        }
+
+        String indexStr = name.substring(openBraceIdx + 1, closeBraceIdx);
+        try {
+            int idx = Integer.valueOf(indexStr);
```
MykolaGolubyev: we either have a util for conversion from a string to number avoiding try-catch or we should add

webtau-http/src/test/groovy/com/twosigma/webtau/http/datanode/StructuredDataNodeTest.groovy
```diff
@@ -123,4 +123,75 @@ class StructuredDataNodeTest {
...
+    @Test
+    void "should access array element via path"() {
+        def node = DataNodeBuilder.fromMap(new DataNodeId("body"), [
+            key1: [[name: 'name1'], [name: 'name2']]
+        ])
+
+        Map<String, String> firstElement = node.get('key1[0]').get()
+        firstElement.should == [name: 'name1']
+
+        String firstName = node.get('key1[0].name').get()
```
MykolaGolubyev: how about case like `root.child[idx].anotherChild`?

webtau-http/src/test/groovy/com/twosigma/webtau/http/datanode/StructuredDataNodeTest.groovy
```diff
@@ -123,4 +123,75 @@ class StructuredDataNodeTest {
...
+        firstName.should == 'name1'
+
+        String lastName = node.get('key1[-1].name').get()
```
MykolaGolubyev: nice

tsiq-karold:
> I think that even in Groovy case it can be useful if someone decides to get/build a path at runtime. Or have a path inside a config file (i.e. different versions of api to access temperature from a city).

That is true, I guess there are Groovy usecases because I think its value to Java users is much more significant.  I also need it for #221 .

webtau-http/src/main/java/com/twosigma/webtau/http/datanode/StructuredDataNode.java
tsiq-karold: That'll throw.  That's what the `additional*Idx` checks are for, they look for brackets after the first one.

webtau-http/src/main/java/com/twosigma/webtau/http/datanode/StructuredDataNode.java
tsiq-karold: OK, I'll check

webtau-http/src/test/groovy/com/twosigma/webtau/http/datanode/StructuredDataNodeTest.groovy
tsiq-karold: You mean using a var for the index?

MykolaGolubyev: no, just a usecase where index is in the middle, not just start or end

webtau-http/src/main/java/com/twosigma/webtau/http/datanode/StructuredDataNode.java
MykolaGolubyev: doh

webtau-http/src/test/groovy/com/twosigma/webtau/http/datanode/StructuredDataNodeTest.groovy
tsiq-karold: Oh, ok.  It should work but I'll add a test.

webtau-http/src/main/java/com/twosigma/webtau/http/datanode/StructuredDataNode.java
```diff
@@ -59,19 +59,66 @@ public DataNodeId id() {
...
+        } catch (NumberFormatException e) {
+            throw new IllegalArgumentException("Requested index " + indexStr + " of name " + name.substring(0, openBraceIdx) + " is not an integer");
+        }
+    }
+
     @Override
-    public boolean has(String name) {
-        return children.containsKey(name);
+    public boolean has(String pathOrName) {
+        return !(get(pathOrName) instanceof NullDataNode);
```
tsiq-karold: @MykolaGolubyev I wasn't massively keen on this approach.  One possible small improvement would be to add an `isNull` on `DataNode` instead of checking the type.

The alternative would require a workflow very similar to the `get` functionality but instead of doing the ultimate `children.get`, it would do `children.containsKey` so it's actually no better in terms of computational effort and would either require lots of dupe code or some ugly parameterisation.

MykolaGolubyev: `isNull` sounds reasonable. Adding default method should be backward compatible right?

tsiq-karold: Yes.

webtau-http/src/main/java/com/twosigma/webtau/http/datanode/StructuredDataNode.java
tsiq-karold: Actually, we don't really need the try-catch here as such, `NumberFormatException` is a runtime exception.  I just wanted a more useful error message which is why I catch it and throw something different.

tsiq-karold:
Created https://github.com/twosigma/webtau/issues/344 to track documentation.

webtau-http/src/main/java/com/twosigma/webtau/http/datanode/StructuredDataNode.java
MykolaGolubyev: I see, I would still extract it somewhere locally.

webtau-http/src/main/java/com/twosigma/webtau/http/datanode/StructuredDataNode.java
```diff
@@ -59,19 +59,66 @@ public DataNodeId id() {
     }
 
     @Override
-    public DataNode get(String name) {
+    public DataNode get(String nameOrPath) {
         if (isList()) {
-            return getAsCollectFromList(name);
+            return getAsCollectFromList(nameOrPath);
```
MykolaGolubyev: now I got confused as to how this will work in case of `path`. `getAsCollectFromList` signature is also talking about `name` only

MykolaGolubyev:
> I left one comment, but up to you.

Not sure now how to un-approve. There is `getAsCollectFromList` that I need your help to understand how will it work in case of `path` not `name`

tsiq-karold:
> > I left one comment, but up to you.
> 
> Not sure now how to un-approve. There is `getAsCollectFromList` that I need your help to understand how will it work in case of `path` not `name`

Conceptually it hasn't really changed.  It'll apply the name or path logic to each member of the list if you reference a list element without `[]`.

PR #349, opened 2019-06-06
prepare for next development iteration

tsiq-clemens:
update 2 poms to the next version like the others


tsiq-clemens:
putting this change in with 351

PR #351, opened 2019-06-06
Use path

tsiq-clemens:
fixes https://github.com/twosigma/webtau/issues/350 by forcing a Path, which works with `Files.exists()`

### DIFF
--- a/webtau-cli-groovy/pom.xml
+++ b/webtau-cli-groovy/pom.xml
@@ -7,7 +7,7 @@
     <parent>
         <groupId>com.twosigma.webtau</groupId>
         <artifactId>webtau-parent</artifactId>
-        <version>1.9-SNAPSHOT</version>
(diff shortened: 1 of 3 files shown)


webtau-cli-groovy/pom.xml
```diff
@@ -7,7 +7,7 @@
     <parent>
         <groupId>com.twosigma.webtau</groupId>
         <artifactId>webtau-parent</artifactId>
-        <version>1.9-SNAPSHOT</version>
+        <version>1.10-SNAPSHOT</version>
```
tsiq-karold: @MykolaGolubyev any ideas why the release:prepare might not have updated these two modules?

MykolaGolubyev: My guess is I did branch and didn’t merge later.

PR #352, opened 2019-06-11
stop webtau feature tests from overriding html reports

### LINKED ISSUE #323: Run tests with custom report names
Right now if you run `mvn test` and some tests failed, their html report will have been overwritten by subsequent tests so it's hard to debug.  The reports generally go into `target` so won't pollute workspaces.

MykolaGolubyev:
Closes #323

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ target
 node_modules
 webtau.*report*.html
 webtau-report-*.html
+*-webtau-report.html
 .webtau.cache.json
 .vertx
 webtau-http-groovy/doc-artifacts/*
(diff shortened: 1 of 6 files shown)


tsiq-clemens:
do you want to have 1 file per test file? or a timestamped combined file per test run?

tsiq-karold:
I don't think we need the timestamp.  Rebuilding will nuke the code in target anyway.  The problem is more about multiple tests in the same run overriding the report.

MykolaGolubyev:
I think reports are being generated next to the test file right now and marked as `ignore` in `git`

PR #357, opened 2019-06-14
Capture HTTP URLs for documentation

### LINKED ISSUE #353: Capture HTTP URL
`http.doc.capture` does not capture the URL invoked.  It would be useful to capture both the relative and absolute URLs.

tsiq-karold:
Fixes #353

### DIFF
--- a/webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
+++ b/webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
@@ -522,7 +522,7 @@ class HttpGroovyTest implements HttpConfiguration {
             header.authorization.should == 'topSecret'
         }
 
-        String artifactName = 'echo-body-and-header'
+        String artifactName = 'echo-body-and-header-redacted'
         http.doc.capture(artifactName)
 
         Path docRoot = DocumentationArtifactsLocation.resolve(artifactName)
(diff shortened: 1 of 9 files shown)


webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
```diff
@@ -522,21 +520,36 @@ class HttpGroovyTest implements HttpConfiguration {
...
         }
 
-        readAndAssertCapturedFile(docRoot, 'response.header.txt') { responseHeaderFile ->
+        readAndAssertCapturedFile(artifactName, 'response.header.txt') { responseHeaderFile ->
             authHeader(responseHeaderFile).should == redactedAuth
         }
     }
 
+    @Test
+    void "urls are captured"() {
```
MykolaGolubyev: lets add `doc` to the title somewhere

webtau-http/src/main/java/com/twosigma/webtau/http/HttpDocumentation.java
```diff
@@ -52,6 +55,11 @@ public void capture() {
             capturePaths();
         }
 
+        private void captureUrl() {
+            FileUtils.writeTextContent(path.resolve("request.url.path.txt"), UrlUtils.extractPath(lastValidationResult.getFullUrl()));
```
MykolaGolubyev: I wonder if we should capture the URL user provided in a test? Does `extractPath` keeps `query` part of the url?

tsiq-karold: That's a great point and no, it won't.  I had considered that initially but the result object the capture looks at does not have the user specified url.  I'll see what it'll take to have it included in there but I agree the distinction should ideally be user specified part vs. full url rather than path vs. url.

One interesting question is what will happen if someone specifies query params but not in the url part.  I guess I'll test and find out 😄

PR #362, opened 2019-06-19
Include relative path of scenario in report

### LINKED ISSUE #347: Include relative path of scenario in report
With the recent change to search for scenarios recursively it's possible to have some confusing entries in the report.  If the scenarios dir looks like this for example:
* foo:
** test.groovy
* bar:
** test.groovy

Then the report will have two sections entitled `test.groovy`.  We should include the relative path.

tsiq-karold:
fixes #347 

This looks like a lot of files but that's mostly because I added another field to be captured in `run-details.json` for feature tests so had to update all the expected files.

This is what the report looks like for the feature test I added:
image


webtau-feature-testing/test-expectations/scenarios/concept/runOnlySelected/run-details.json
```diff
@@ -1,15 +1,19 @@
 {
   "scenarioDetails" : [ {
     "scenario" : "step one",
+    "shortContainerId" : "runOnlySelected.groovy",
```
MykolaGolubyev: I understand the purpose and how it tested your scenario, but I don't like having them in every expectation very much. Don't have an alternative yet.

webtau-groovy/src/main/groovy/com/twosigma/webtau/cfg/WebTauGroovyCliArgsConfigHandler.groovy
```diff
@@ -60,17 +61,24 @@ class WebTauGroovyCliArgsConfigHandler implements WebTauConfigHandler {
...
 
-    List<Path> testFilesWithFullPath() {
+    List<TestFile> testFilesWithFullPath() {
         return argsConfig.testFiles.collectMany { fileName ->
             def path = cfg.workingDir.resolve(Paths.get(fileName)).toAbsolutePath()
-            def paths = []
+            def testFiles = []
             if (Files.isDirectory(path)) {
-                path.toFile().eachFileRecurse(FileType.FILES) { paths << it.toPath() }
+                int baseDirEndIdx = path.toString().length()
```
MykolaGolubyev: it is getting a bit of nesting, should we move it to a separate private method?

webtau-reactjs/src/report/Report.js
```diff
@@ -247,9 +247,13 @@ function fullContainerId(test) {
...
 
 function shortContainerId(test) {
-    return test.className ?
-        shortenClassName(test.className):
-        shortenFileName(test.fileName)
+    if (test.shortContainerId) {
+        return test.shortContainerId
+    } else if (test.className) {
+        return shortenClassName(test.className)
+    } else {
```
MykolaGolubyev: will we ever have this case?

webtau-standalone-runner-groovy/src/main/groovy/com/twosigma/webtau/runner/standalone/StandaloneTestRunner.groovy
```diff
@@ -46,12 +48,14 @@ class StandaloneTestRunner {
         this.isTerminated = new AtomicBoolean(false)
     }
 
-    void process(Path scriptPath, delegate) {
+    void process(TestFile testFile, delegate) {
+        Path scriptPath = testFile.path
         currentTestPath = scriptPath.isAbsolute() ? scriptPath : workingDir.resolve(scriptPath)
+        currentShortContainerId = testFile.shortContainerId
```
MykolaGolubyev: containerId makes sense in html report context. Not sure if it is informative enough in TestFile context

webtau-feature-testing/test-expectations/scenarios/concept/runOnlySelected/run-details.json
tsiq-karold: Tell me about it, I didn't really want to have to modify all these files! :)

One alternative would be to allow feature test to pass an optional set of fields to capture from the `StandaloneTest` in addition to the default scenario and stepsSummary.

webtau-reactjs/src/report/Report.js
tsiq-karold: Shouldn't do.  I had initially left it in while I was working on this change but I think this can go now.

webtau-standalone-runner-groovy/src/main/groovy/com/twosigma/webtau/runner/standalone/StandaloneTestRunner.groovy
tsiq-karold: Yeah, so I initially had it as shortFileName in TestFile and then it was changed into shortContainerId when throwing stuff over to JS.  I then decided to make it consistent across the board.  Happy to go back, not sure I feel strongly about it either way.

tsiq-karold:
That test that fails in Travis passes locally 😢

tsiq-karold:
@MykolaGolubyev the test that fails in Travis passes locally.  It looks to me like the order of file traversal somehow differs between my Mac and Travis' linux boxes, either an OS or JVM version difference.  One possible fix would be to deserialize `scenarioDetails` into a set rather than a list or sort it.  Any other ideas/thoughts?

MykolaGolubyev:
Scenarios themselves should be a list since order must be guaranteed. Is the problem with files resolved in different order?

tsiq-karold:
> Is the problem with files resolved in different order?

Yes.  Maybe a better solution would be to sort the files we discover?

MykolaGolubyev:
@tsiq-karold sorting files may be not idea as there could be a reason user wants a specific order without using `*`. You already have `containerId` captured, so maybe we can sort by that before comparison?

webtau-feature-testing/src/main/groovy/com/twosigma/webtau/featuretesting/WebTauEndToEndTestRunner.groovy
```diff
@@ -129,6 +129,16 @@ class WebTauEndToEndTestRunner implements StepReporter, StandaloneTestListener {
         }
     }
 
+    private static Map sortTestDetailsByContainerId(Map testDetails) {
+        def scenarioDetails = testDetails.scenarioDetails
```
MykolaGolubyev: should we make a copy of list or explicitly modify in place? RIght now it makes it look like it returns a copy, but it doesn't, does it?

tsiq-karold: Bit of both :)  It creates a sorted copy of the list but then overrides it in the main map.  I return the map just to make the calling code easier.

Having said that, I can see why it's confusing.  I'll make it all copy.

Issue #369, opened 2019-06-27
CLI - documentation


tsiq-karold:
Some things to remember:
* resolution info - current working dir, setting path, etc.

PR #370, opened 2019-06-27
Cli path support

### LINKED ISSUE #359: Cli scripts location resolution
When you run your script (as opposite to a system command like `curl`), there should be a script location resolution.

Open questions:
* how to decide whether to `resolve` a given command/scrip or not. one possible solution is to try to resolve and if it works, use resolved version.
* how to specify resolution locations. Ideally it should be an array of locations, but we need to be able to override the value through command line. Should it be a `;` delimited string?

tsiq-karold:
fixes #359 

Some documentation will be required for this which I have noted down in #369 .

First thing I tried is simply modifying the PATH that the ProcessBuilder has.  Unfortunately, this did not work.  As far as I can tell, that modifies only the environment of the subprocess and does not impact resolution of the script/command itself.  There are a number of SO articles that back up this theory and don't butcher the terminology as much as I may be doing 😁 
https://stackoverflow.com/questions/26992165/how-to-set-path-environment-variable-in-processbuilder-java-in-windows
https://stackoverflow.com/questions/10035383/setting-the-environment-for-processbuilder

One beneficial side effect is this allows us to delegate quoting (and escaping) and breaking up of individual args to the shell rather than having to do it ourselves.  In fact, I'm tempted to remove our code for doing that altogether and have the non-PATH branch also use bash -c.


MykolaGolubyev:
Thanks for the detailed explanation. Are there guarantees for `bash` presence?

tsiq-karold:
> Thanks for the detailed explanation. Are there guarantees for `bash` presence?

Nope, that's an excellent point.  I'll switch to bin/sh.

webtau-cli-groovy/src/test/groovy/com/twosigma/webtau/cli/CliGroovyTest.groovy
```diff
@@ -31,7 +31,7 @@ class CliGroovyTest {
 
                 output.should == ~/hello/
                 output.should contain('world')
-                output.should contain('"message to world"')
+                output.should contain('message to world')
```
tsiq-karold: I believe the previous behaviour was actually incorrect.  Invoking the same command from a terminal did not result in quoted message to world:
```bash
$ scripts/hello "message to world"
hello world message to world
line in the middle
more text
error line one
error line two
```

PR #379, opened 2019-07-11
Webtau uber jar + shading attempt

MykolaGolubyev:
Not working yet. Current problem is I shade groovy as well and even though I use transformations, it doesn't shade properly `META-INF/groovy/org.codehaus.groovy.runtime.ExtensionModule`. I think it is because it is not in a standard `services` dir. 

Should we even shade Groovy? One of the reason I want to shade is to mitigate monorepo deps problems. Alternatively I could separate into groovy and non groovy jars.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -127,6 +127,7 @@
         <module>webtau-maven-plugin-test</module>
         <module>webtau-maven-plugin-test-skip</module>
         <module>webtau-report-testing</module>
+        <module>webtau-shaded</module>
         <module>webtau-dist</module>
         <module>webtau</module>
     </modules>
--- a/webtau-dist/pom.xml
+++ b/webtau-dist/pom.xml
@@ -41,6 +41,12 @@
                 </exclusion>
             </exclusions>
         </dependency>
+
+        <dependency>
+            <groupId>org.codehaus.groovy</groupId>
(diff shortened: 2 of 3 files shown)


tsiq-karold:
What's the purpose of the uber jar + shading?

MykolaGolubyev:
Shading to be able to use as a dependency in a monorepo setup without fighting conflicts and versions mismatch.

tsiq-karold:
I guess it depends on how your monorepo/build is setup and I can see how it would definitely be an issue if using the junit runners but if you've just got groovy scenarios which you invoke via the Maven plugin, this shouldn't be an issue?

MykolaGolubyev:
@tsiq-karold one of the usecases is to wrap opensource version and provide some of the global defaults for config values like chromeLocation. In order to do that we need to have jars as a dependency. 
That is on top of using webtau with junit runners for unit like tests.

tsiq-karold:
OK, makes sense.

> Should we even shade Groovy

Do you mean webtau groovy code?  Why would we shade webtau?

MykolaGolubyev:
Shade the groovy itself as a dependency. Vs split into Uber jar and groovy jar. So then when we bring webtau internally we will bring only Uber jar and use internal version of groovy.

PR #388, opened 2019-07-24
reduce groovy deps footprint

### LINKED ISSUE #387: Replace groovy-all with explicit dependencies
We use `groovy-all.pom` as a groovy dependency. It comes with a lot of packages and we don't use most of them.
We need to explicitly specify only the dependencies we care about.

MykolaGolubyev:
closes #387

### DIFF
--- a/webtau-cli-groovy/pom.xml
+++ b/webtau-cli-groovy/pom.xml
@@ -25,8 +25,14 @@
 
         <dependency>
             <groupId>org.codehaus.groovy</groupId>
-            <artifactId>groovy-all</artifactId>
-            <type>pom</type>
+            <artifactId>groovy-groovydoc</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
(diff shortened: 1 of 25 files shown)


webtau-cli-groovy/pom.xml
```diff
@@ -25,8 +25,14 @@
...
         <dependency>
             <groupId>org.codehaus.groovy</groupId>
-            <artifactId>groovy-all</artifactId>
-            <type>pom</type>
+            <artifactId>groovy-groovydoc</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>junit</groupId>
```
tsiq-clemens: Junit 4?

MykolaGolubyev: yeah, most of the tests in webtau to test webtau are in junit4.

PR #393, opened 2019-07-31
add --example option to bootstrap examples ready to be ran

MykolaGolubyev:
Idea is users can just do `webtau --example` and run generated examples without any setup to get the feel for the tool.
Folks who wrap webtau can register their own examples that are, say, company specific.

There is a new module `webtau-cli-testing` that is a placeholder for now.
Need to figure out how to make webtau cli testing capabilities to test webtau cli itself (unzipped dist after assembly).

### DIFF
--- a/webtau-groovy/src/main/groovy/com/twosigma/webtau/cfg/WebTauCliArgsConfig.groovy
+++ b/webtau-groovy/src/main/groovy/com/twosigma/webtau/cfg/WebTauCliArgsConfig.groovy
@@ -16,14 +16,19 @@
 
 package com.twosigma.webtau.cfg
 
+import com.twosigma.webtau.examples.ExamplesScaffolder
 import org.apache.commons.cli.CommandLine
 import org.apache.commons.cli.DefaultParser
 import org.apache.commons.cli.HelpFormatter
 import org.apache.commons.cli.Options
 import org.apache.commons.cli.ParseException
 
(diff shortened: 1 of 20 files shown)


webtau-groovy/src/main/groovy/com/twosigma/webtau/cfg/WebTauCliArgsConfig.groovy
```diff
@@ -62,16 +67,26 @@ class WebTauCliArgsConfig {
...
+            scaffoldExamples()
+        } else if (commandLine.hasOption(HELP_OPTION) || commandLine.argList.isEmpty()) {
+            printHelp(options)
+        } else {
+            testFiles = new ArrayList<>(commandLine.argList)
         }
+    }
+
+    private void scaffoldExamples() {
+        ExamplesScaffolder.scaffold(Paths.get(""))
```
tsiq-karold: Do we need the argument here?  We only ever seem to pass `Paths.get("")`.  If we do need it but this is a sensible default then I'd suggest adding an argument-less overload of `scaffold`.

MykolaGolubyev: I was entertaining the possibility of handling `--workingDir` value in future

PR #396, opened 2019-08-01
Support json derived content types

### LINKED ISSUE #372: Webtau does not handle application/problem+json content type
There is a draft spec for this content type: https://tools.ietf.org/html/draft-nottingham-http-problem-07 and I've unfortunately unearthed an API that uses it.  Webtau really doesn't like it.

The body of such a response is for all intents and purposes the same as `application/json` but webtau treats it somewhat as binary.  The CLI output shows the response as `[binary content]` and if I try and expand the relevant row in the `HTTP calls` tab, the report goes blank and the console shows this error:
```
VM75 webtau.report.html:14 SyntaxError: Unexpected token b in JSON at position 1
    at JSON.parse (<anonymous>)
    at r (VM75 webtau.report.html:14)
    at sn (VM75 webtau.report.html:14)
    at In (VM75 webtau.report.html:14)
    at _r (VM75 webtau.report.html:14)
    at kr (VM75 webtau.report.html:14)
    at Hr (VM75 webtau.report.html:14)
    at Yr (VM75 webtau.report.html:14)
    at $r (VM75 webtau.report.html:14)
    at Ue (VM75 webtau.report.html:14)
```

tsiq-karold:
closes #372


webtau-http/src/main/java/com/twosigma/webtau/http/HttpResponse.java
```diff
@@ -70,7 +70,7 @@ public boolean hasContent() {
     }
 
     public boolean isJson() {
-        return contentType.contains("/json");
+        return contentType.matches(".*/.*\\+?json");
```
MykolaGolubyev: wonder if adding regepx now warrants a unit test to test negative cases

tsiq-karold: What sort of negative test did you have in mind?  Something like responding with `application/notreallyjson`?

I was in two minds whether to do this via regex to be honest.  I wondered about `endsWith(“json”)` but that would also accept things like `notreallyjson` which I don’t think is correct.  Also considered `contains(“/json”) || contains (“+json”)`.  I chose regex because it felt more correct but I could probably be convinced otherwise.

MykolaGolubyev: Yes, something like the example you have.

tsiq-karold: added

Issue #227, opened 2018-12-12
Implement equals and compareTo on DataNode


tsiq-karold:
This will allow things like:
```
if (foo.bar == "blah")
```
on responses for example.

MykolaGolubyev:
This turned out to be tricky in Groovy :(
if it is only `==` then no issues, but for symmetry we may need to support `foo.bar > 10` and this leads to complications
https://stackoverflow.com/questions/53798075/groovy-compareto-for-customclass-and-numbers-strings

MykolaGolubyev:
@tsiq-karold any ideas? we added `.get()` so not that much typing, the issue is there is no compile time warning. Wonder if some sort of AST should come to the rescue. I.e. detect that value inside if is an actual DataNode (at compile time) and automatically add .get() to it?

tsiq-karold:
We could do the AST trick, not sure it's necessarily worth the effort though.

MykolaGolubyev:
Groovy 2.5.7 changelog mentioned some fixes around the symmetry. Unfortunately I nuked the branch. Will try again. 
https://issues.apache.org/jira/browse/GROOVY-7954

MykolaGolubyev:
There is still an issue with handling `==` comparison of objects that don't implement `Comparable` (e.g. arrays).

MykolaGolubyev:
closing this for now as not much else can be done

PR #399, opened 2019-08-06
DataNode if-else comparison capabilities

MykolaGolubyev:
Addresses #227 for the most part.
The complex types are still not properly handled due to Groovy impl details.


webtau-core/src/main/java/com/twosigma/webtau/data/traceable/TraceableValue.java
```diff
@@ -44,6 +45,8 @@ public void updateCheckLevel(CheckLevel newCheckLevel) {
             return;
         }
 
+        newCheckLevel = isAlwaysFuzzyPassedTracing.get() ? CheckLevel.FuzzyPassed : newCheckLevel;
```
tsiq-karold: Not generally a fan of modifying parameters.  Can we change this to use/create a new var?

MykolaGolubyev: Makes sense. Also removed unused modifier lurking there.

PR #400, opened 2019-08-06
add version override extension for custom wrappers

MykolaGolubyev:


MykolaGolubyev:
@tsiq-karold I need to merge this now for internal wrapper org. Please feel free to comment, I will patch based on comments.

tsiq-karold:
> @tsiq-karold I need to merge this now for internal wrapper org. Please feel free to comment, I will patch based on comments.

Looks sensible though I'd argue using the service loader mechanism for this purpose might be overkill, a system property or similar might be more appropriate.

MykolaGolubyev:
@tsiq-karold in my case I need to have access to the original value to modify it: replace snapshot with a company sufix and a build date

PR #429, opened 2019-08-28
Allow programmatically setting Open API spec URL

### LINKED ISSUE #420: Allow programmatically setting configuration options
I'd like to enable programmatic access to setting the configuration options. This will make it easier and cleaner to set the config options with internal TS tooling, where env files are not always supported and to allow us to specify these options the way configuration is commonly specified in Java codebases, e.g. through non-env property files.

mattmetlis:
Fixes #420

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 .idea
 .mvn
 build
+doc-artifacts
 target
 node_modules
 webtau.*report*.html
--- a/webtau-open-api/src/main/java/com/twosigma/webtau/openapi/OpenApiSpecConfig.java
+++ b/webtau-open-api/src/main/java/com/twosigma/webtau/openapi/OpenApiSpecConfig.java
@@ -25,6 +25,9 @@ import java.util.stream.Stream;
 import static com.twosigma.webtau.cfg.ConfigValue.declare;
 import static com.twosigma.webtau.cfg.WebTauConfig.getCfg;
 
+/**
+ * Configuration of the Open API specification location.
+ */
 public class OpenApiSpecConfig implements WebTauConfigHandler {
     static final ConfigValue specUrl = declare("openApiSpecUrl",
             "url of OpenAPI 2 spec against which to validate http calls", () -> "");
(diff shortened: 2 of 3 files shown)


mattmetlis:
> if the intention is only to set the value I would suggest to expose only `setSpecUrl` similarly to how we expose `setBaseUrl`

I'm fine with that, but I'm curious, what's the case where you'd want to expose this kind of getter instead of or in addition to a setter?

MykolaGolubyev:
One example is extensions so they can specify a sourceId of what have overridden it. 
Another is for tests to validate some internals. 
Now that I said this out loud, feel free to provide both a getter and a direct setter.

mattmetlis:
I've added a direct setter as well, and a basic unit test.

mattmetlis:
I notice that the fields on OpenApiSpecConfig are static, whereas on WebTauConfig they are not. This makes OpenApiSpecConfig more difficult to test; for example I'd like to add another unit test to the same test class I just added, but because things are static the tests impact each other. What do you think about making OpenApiSpecConfig more like WebTauConfig in terms of fields being non-static so you can work with an instance?

MykolaGolubyev:
@mattmetlis oh sure, go for it!

webtau-open-api/src/main/java/com/twosigma/webtau/openapi/OpenApiSpecConfig.java
```diff
@@ -34,6 +37,14 @@
 
     private static String fullPath;
 
+    public static ConfigValue getSpecUrl() {
```
MykolaGolubyev: we probably should rename this to `getSpecUrlConfigValue` to be consistent with the other values exposed this way.

.gitignore
```diff
@@ -2,6 +2,7 @@
 .idea
 .mvn
 build
+doc-artifacts
```
MykolaGolubyev: interesting, do we still have places were we generate `doc-artifacts` outside of `target` dir?

MykolaGolubyev: @mattmetlis do you know if anything else is pending to merge this?

MykolaGolubyev:
Not an issue at the moment

PR #431, opened 2019-09-11
add ran ratio vs skipped. adjust math.

### LINKED ISSUE #382: Skipped tests show up as tests with problems in report
If there are tests which are skipped due to `onlyForEnv` for example, they count towards "Total with Problems" in the HTML report.  This is a bit misleading.

Screenshot 2019-07-12 at 10 29 45
Screenshot 2019-07-12 at 10 29 59

MykolaGolubyev:
adds new section to report summary with ran ratio
image

also should close #382


MykolaGolubyev:
@tsiq-karold could you please review the fix?

PR #449, opened 2019-10-28
Matchers as a separate class. Ddjt->WebTauCore. deprecate beGreater/Less

MykolaGolubyev:
* `Matchers` as a separate class
* rename `Ddjt` to `WebTauCore`
* deprecate `beGreater/Less` family in favor of `shouldBe`

Note that existing class `Ddjt` is gone. It was not supposed to be used directly for `http` and `web ui` testing. 
In theory it is a backward incompatible change, but in practice it should be fine. 
I am reluctant to roll `2.x` version yet. 

Any thoughts @tsiq-clemens, @tsiq-karold?

### DIFF
--- a/webtau-core/src/main/java/com/twosigma/webtau/Matchers.java
+++ b/webtau-core/src/main/java/com/twosigma/webtau/Matchers.java
@@ -67,7 +67,8 @@ public class Matchers {
     /**
      * Contain matcher
      * <pre>
-     * actual(value).should(contain(10));
+     * actual(collection).should(contain(10));
+     * actual(text).should(contain("hello"));
      * </pre>
      * @param expected value to be contained
      * @return matcher instance
@@ -77,9 +78,9 @@ public class Matchers {
     }
 
     /**
-     * Contain matcher. Alias to contain
+     * Containing matcher. Alias to contain
      * <pre>
-     * actual(value).should(containing(10));
+     * actual(collectionWithText).should(contain(containing("hello")));
      * </pre>
      * @param expected value to be contained
      * @return matcher instance
--- a/webtau-core/src/main/java/com/twosigma/webtau/Matchers.java
+++ b/webtau-core/src/main/java/com/twosigma/webtau/Matchers.java
@@ -1,0 +1,268 @@
+package com.twosigma.webtau;
+
+import com.twosigma.webtau.expectation.*;
+import com.twosigma.webtau.expectation.code.ThrowExceptionMatcher;
+import com.twosigma.webtau.expectation.contain.ContainMatcher;
+import com.twosigma.webtau.expectation.equality.*;
+
+import java.util.regex.Pattern;
+
+/**
+ * Convenient place to discover all the available matchers
+ */
+public class Matchers {
+    /**
+     * Starting point of a value matcher
+     * <pre>
+     * actual(value).should(beGreaterThan(10));
+     * actual(value).shouldNot(beGreaterThan(10));
+     * </pre>
+     * Note: In Groovy you can just do <code>value.should beGreaterThan(10)</code>
+     * @param actual value to assert against
+     * @return Object to chain a matcher against
+     */
+    public static ActualValueExpectations actual(Object actual) {
+        return new ActualValue(actual);
+    }
+
+    /**
+     * Starting point of a code matcher
+     * <pre>
+     * code(() -> {
+     *    businessLogic(-10);
+     * }).should(throwException(IllegalArgumentException.class, "negative are not allowed"));
+     * </pre>
+     *
+     * @param codeBlock code to match against
+     * @return Object to chain a matcher against
+     */
+    public static ActualCodeExpectations code(CodeBlock codeBlock) {
+        return new ActualCode(codeBlock);
+    }
+
+    /**
+     * Equal matcher
+     * <pre>
+     * actual(value).should(equal(10));
+     * </pre>
+     * @param expected value to be equal to
+     * @return matcher instance
+     */
+    public static EqualMatcher equal(Object expected) {
+        return new EqualMatcher(expected);
+    }
+
+    /**
+     * Not equal matcher
+     * <pre>
+     * actual(value).should(notEqual(10));
+     * </pre>
+     * @param expected value to not be equal to
+     * @return matcher instance
+     */
+    public static NotEqualMatcher notEqual(Object expected) {
+        return new NotEqualMatcher(expected);
+    }
+
+    /**
+     * Contain matcher
+     * <pre>
+     * actual(value).should(contain(10));
+     * </pre>
+     * @param expected value to be contained
+     * @return matcher instance
+     */
+    public static ContainMatcher contain(Object expected) {
+        return new ContainMatcher(expected);
+    }
+
+    /**
+     * Contain matcher. Alias to contain
+     * <pre>
+     * actual(value).should(containing(10));
+     * </pre>
+     * @param expected value to be contained
+     * @return matcher instance
+     */
+    public static ContainMatcher containing(Object expected) {
+        return new ContainMatcher(expected);
+    }
+
+    /**
+     * Greater than matcher
+     * <pre>
+     * actual(value).shouldBe(greaterThan(10));
+     * </pre>
+     * @param expected value to be greater than
+     * @return matcher instance
+     */
+    public static GreaterThanMatcher greaterThan(Object expected) {
+        return new GreaterThanMatcher(expected);
+    }
+
+    /**
+     * Greater than or equal matcher
+     * <pre>
+     * actual(value).shouldBe(greaterThanOrEqual(10));
+     * </pre>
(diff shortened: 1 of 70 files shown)


webtau-core/src/main/java/com/twosigma/webtau/Matchers.java
```diff
@@ -0,0 +1,268 @@
...
+     * </pre>
+     * @param expected value to be contained
+     * @return matcher instance
+     */
+    public static ContainMatcher contain(Object expected) {
+        return new ContainMatcher(expected);
+    }
+
+    /**
+     * Contain matcher. Alias to contain
```
tsiq-karold: `Containing matcher` rather than `Contain matcher` I think.

webtau-core/src/main/java/com/twosigma/webtau/Matchers.java
```diff
@@ -0,0 +1,268 @@
...
+     * @return matcher instance
+     */
+    public static NotEqualMatcher notEqual(Object expected) {
+        return new NotEqualMatcher(expected);
+    }
+
+    /**
+     * Contain matcher
+     * <pre>
+     * actual(value).should(contain(10));
```
tsiq-karold: How about `actual(containValue).should(contain(10));` just to make it clearer that this is a matcher for collections.

webtau-core/src/main/java/com/twosigma/webtau/Matchers.java
```diff
@@ -0,0 +1,268 @@
...
+    public static LessThanOrEqualMatcher lessThanOrEqual(Object expected) {
+        return new LessThanOrEqualMatcher(expected);
+    }
+
+    /**
+     * Throw exception <code>code</code> matcher.
+     * <pre>
+     * code(() -> {
+     *     businessLogic(-10);
+     * }).should(throwException("negative are not allowed"));
```
tsiq-karold: This should be either `negatives are not allowed` or better still `negative numbers`.

webtau-core/src/main/java/com/twosigma/webtau/Matchers.java
```diff
@@ -0,0 +1,268 @@
...
+     * code(() -> {
+     *      businessLogic(-10);
+     * }).should(throwException(Pattern.compile("negative .* not allowed")));
+     * </pre>
+     * @see #code(CodeBlock)
+     *
+     * @param expectedMessageRegexp regular pattern to match expected exception message
+     * @return matcher instance
+     */
+    public static ThrowExceptionMatcher throwException(Pattern expectedMessageRegexp) {
```
tsiq-karold: Should we add an alias to this one which takes a string and is called something like `throwExceptionWithMessageMatching`?

webtau-core/src/main/java/com/twosigma/webtau/WebTauCore.java
```diff
@@ -43,11 +43,9 @@
 import java.util.regex.Pattern;
 
 /**
- * Data Driven Java Testing
- * <p>
- * Convenient class for a single static * imports
+ * Convenient class for a single static * imports to have matchers and helper functions available for your test
  */
-public class Ddjt {
+public class WebTauCore extends Matchers {
```
tsiq-karold: Are we ok with this back compat break?

tsiq-karold: I've re-read the description of the PR.  I'm OK with this break I think because I doubt anyone other than Clemens is using it.

webtau-core/src/main/java/com/twosigma/webtau/Matchers.java
MykolaGolubyev: not sure what `containValue` adds here. I do agree that we need to show that it works for containers. `contain` works for strings to.

maybe something like this?
```
actual(collection).should(contain(10)); 
actual(text).should(contain("hello")); 
```

webtau-core/src/main/java/com/twosigma/webtau/Matchers.java
MykolaGolubyev: Have mixed feelings about this one. May start going into the direction of multiple overloads like junit does: `assertEquals`, `assertEqualsArrays`, `assertEqualsNumbers`. 
It is readable on the client side, at least Groovy portion: 
```
throwException(~/negative/)
```
But as you pointed before, it could be just because I know how it works.

webtau-core/src/main/java/com/twosigma/webtau/Matchers.java
MykolaGolubyev: sounds good, thank you

webtau-core/src/main/java/com/twosigma/webtau/Matchers.java
MykolaGolubyev: duh

webtau-core/src/main/java/com/twosigma/webtau/Matchers.java
tsiq-karold: Typo, brain to finger disconnect.

I meant `collectionValue` not `containValue`.

webtau-core/src/main/java/com/twosigma/webtau/Matchers.java
tsiq-karold: I'm happy to leave it out until someone asks for it.

webtau-core/src/main/java/com/twosigma/webtau/Matchers.java
```diff
@@ -0,0 +1,269 @@
...
+     * @return matcher instance
+     */
+    public static ContainMatcher contain(Object expected) {
+        return new ContainMatcher(expected);
+    }
+
+    /**
+     * Containing matcher. Alias to contain
+     * <pre>
+     * actual(collectionWithText).should(contain(containing("hello")));
```
tsiq-clemens: did you mean `actual(collectionWithText).should(containing("hello"));` ?

tsiq-clemens: and since this is now very strange English when you read it, maybe update the example to something where `containing` reads better than `contain`

MykolaGolubyev: “Contain containing” would find an element in an array that contains specified text

webtau-core/src/main/java/com/twosigma/webtau/Matchers.java
tsiq-clemens: I think the overlo[ar]ds are sufficiently expressive, but then again I only use this with groovy.

webtau-core/src/main/java/com/twosigma/webtau/Matchers.java
```diff
@@ -0,0 +1,269 @@
...
+    public static ThrowExceptionMatcher throwException(Class expectedClass, Pattern expectedMessageRegexp) {
+        return new ThrowExceptionMatcher(expectedClass, expectedMessageRegexp);
+    }
+
+    /**
+     * Throw exception <code>code</code> matcher.
+     * <pre>
+     * code(() -> {
+     *      businessLogic(-10);
+     * }).should(throwException(IllegalArgumentException.class, "negative are not allowed"));
```
tsiq-clemens: same as Karol's comment above on the error message

webtau-core/src/main/java/com/twosigma/webtau/Matchers.java
```diff
@@ -0,0 +1,269 @@
...
+     * @see #greaterThan(Object)
+     */
+    @Deprecated
+    public static GreaterThanMatcher beGreaterThan(Object expected) {
+        return greaterThan(expected);
+    }
+
+    /**
+     * Deprecated. Instead use
+     * <pre>
```
tsiq-clemens: you can use the `@deprecated ` javadoc annotation

webtau-core/src/main/java/com/twosigma/webtau/WebTauCore.java
tsiq-clemens: I can't guarantee that, I've been showing DDJT to others -- just mention it in the release notes / readme as well as suggested (non-)usage.

webtau-core/src/test/java/com/twosigma/webtau/expectation/code/ThrowExceptionMatcherTest.java
```diff
@@ -20,12 +20,16 @@
...
-import static com.twosigma.webtau.Ddjt.throwException;
+import static com.twosigma.webtau.WebTauCore.code;
+import static com.twosigma.webtau.WebTauCore.throwException;
 
 public class ThrowExceptionMatcherTest {
     @Test
     public void examples() {
+        code(() -> {
+            businessLogic(-10);
+        }).should(throwException("negative are not allowed"));
```
tsiq-clemens: see above comments

PR #458, opened 2019-11-24
http.delete queryParams overload

### LINKED ISSUE #336: Http.delete has no overrides that allow specifying query params

MykolaGolubyev:
closes #336


webtau-http/src/test/java/com/twosigma/webtau/http/HttpOverloadsTestCommon.java
```diff
@@ -39,6 +40,7 @@
     public static final String PATH_EXPECTED_RETURN = "/full-echo";
 
     public static final HttpQueryParams query = http.query("a", "1", "b", "text");
+    public static final Map<String, ?> queryAsMap = WebTauCore.aMapOf("a", "1", "b", "text");
```
tsiq-karold: I suspect there are existing tests that could make use of this, specifically tests for query param overrides for other http methods.

MykolaGolubyev: Not sure I follow, could you please clarify?

tsiq-karold: I'm saying things like https://github.com/twosigma/webtau/blob/master/webtau-http/src/test/java/com/twosigma/webtau/http/HttpJavaTest.java#L236 could use the `queryAsMap` you've defined here instead of redefining their own.

tsiq-karold: OK, as discussed, these are completely unrelated classes (I assumed one was a base class of the other) so approving it as is.

PR #460, opened 2019-11-25
fix config cleanup in between multiple maven executions

### LINKED ISSUE #426: enable sequential runs with different config files
use case: I have a set of groovy tests to run against v1 of a swagger API with a docPath of test-artifacts/v1, and a different set of tests to run against v2 of the API, with docPath test-artifacts/v2.
I need to run those tests against different webtau config files.
If I run the 2 suites with the webtau-maven-plugin (using the iterator-maven plugin as wrapper), the 2nd run acknowledges the different config file and test files on the console, but does not apply any of its values for docPath, url, apiSpec, etc.

MykolaGolubyev:
closes #426


tsiq-clemens:
nice, thanks!

PR #461, opened 2019-11-27
Http ref doc

MykolaGolubyev:


webtau-docs/webtau/REST/HTTP-calls.md
```diff
@@ -0,0 +1,75 @@
...
+| PUT                | 200                      |
+| PUT (no content)   | 204                      |
+| DELETE             | 200                      |
+| DELETE (no content)| 204                      |
+
+# GET
+
+:include-java: com/twosigma/webtau/http/Http.java {entry: "get(String, HttpQueryParams, HttpHeader, HttpResponseValidatorWithReturn)", signatureOnly: true, title: "full signature"}
+
+```columns {left: {portion: 4, align: "right"}, border: true}
```
MykolaGolubyev: This needs to be replaced with a better `znai` implementation for columns with multiple entries. Right now layout can go wild.

webtau-docs/webtau/REST/HTTP-calls.md
```diff
@@ -0,0 +1,75 @@
...
+Groovy: 
+:include-groovy: com/twosigma/webtau/http/HttpGroovyOverloadsTest.groovy {entry: "get no query return syntax example", bodyOnly: true, title: "without query"}
+
+Java: 
+:include-java: com/twosigma/webtau/http/HttpJavaOverloadsTest.java {entry: "getNoQueryReturnSyntaxExample", bodyOnly: true, title: "without query"}
+``` 
+
+# Implicit statusCode Check 
+
+If your code omits validation block, or if validation block doesn't assert `statusCode` an implicit status code validation
```
tsiq-karold:
```suggestion
If your code omits the validation block, or if validation block doesn't assert on the `statusCode`, then an implicit status code validation will be performed.
```

webtau-docs/webtau/REST/HTTP-calls.md
```diff
@@ -0,0 +1,75 @@
...
+be performed.
```
tsiq-karold:
```suggestion
```

webtau-docs/webtau/REST/HTTP-calls.md
```diff
@@ -0,0 +1,75 @@
...
+
+If your code omits validation block, or if validation block doesn't assert `statusCode` an implicit status code validation
+be performed. 
+
+| Method             | Implicitly Expected Code |
+| ------------------ |--------------------------|
+| GET                | 200                      |
+| POST               | 201                      |
+| PUT                | 200                      |
+| PUT (no content)   | 204                      |
```
tsiq-karold: Maybe clarify with `no response content`?  For delete this isn't particularly ambiguous given that deletes don't have requests but for put this is arguably slightly ambiguous.

webtau-docs/webtau/REST/HTTP-calls.md
```diff
@@ -0,0 +1,75 @@
...
+
+Java: 
+:include-java: com/twosigma/webtau/http/HttpJavaOverloadsTest.java {entry: "getFullReturnSyntaxExample", bodyOnly: true, title: "full signature with query, header and validation block", commentsType: "inline"}
+:include-java: com/twosigma/webtau/http/HttpJavaOverloadsTest.java {entry: "getNoHeaderReturnSyntaxExample", bodyOnly: true, title: "without header"}
+:include-java: com/twosigma/webtau/http/HttpJavaOverloadsTest.java {entry: "getNoQueryReturnSyntaxExample", bodyOnly: true, title: "without query"}
+:include-java: com/twosigma/webtau/http/HttpJavaOverloadsTest.java {entry: "getOnlyValidationReturnSyntaxExample", bodyOnly: true, title: "only validation"}
+:include-java: com/twosigma/webtau/http/HttpJavaOverloadsTest.java {entry: "getOnlySyntaxExample", bodyOnly: true, title: "only call"}
+``` 
+
+:include-groovy: com/twosigma/webtau/http/HttpGroovyOverloadsTest.groovy {entry: "get full return syntax example", bodyOnly: true, title: "full signature with query, header and validation block", commentsType: "inline"}
```
tsiq-karold: Is this intentionally not in the tabs block?

webtau-http/src/main/java/com/twosigma/webtau/http/HttpHeader.java
```diff
@@ -24,6 +24,9 @@
 
 import static java.util.stream.Collectors.joining;
 
+/**
+ * Represents request and response <code>HTTP</code> header
```
tsiq-karold: While this comment makes sense here, it does produce some confusing documentation in your screenshot because the header param to, for example, `http.get` is annotated as `Represents request and response HTTP header` which for a param of that method is not correct and confusing.

webtau-docs/webtau/REST/HTTP-calls.md
MykolaGolubyev: I probably meant to say "will be performed". Your suggest to remove "be performed" part?

webtau-docs/webtau/REST/HTTP-calls.md
MykolaGolubyev: makes sense

webtau-docs/webtau/REST/HTTP-calls.md
MykolaGolubyev: left over to overcome znai bug (doesn't refresh page if the include-file is inside tab and file is changed :( need to fix that one), will remove.

webtau-http/src/main/java/com/twosigma/webtau/http/HttpHeader.java
MykolaGolubyev: Idea was to describe type and not the parameter. But I see how it can be confused in the context of it usage. 
So I wonder if I should implement znai code reference feature first. I.e. if you hover over `HttpHeader` within the code it will popup its java doc, and if you click then it will navigate you to the reference page.

webtau-docs/webtau/REST/HTTP-calls.md
tsiq-karold: This suggestion in isolation makes no sense.  I just moved the `be performed` part into the suggestion on the line above so deleted this line.

tsiq-karold: It was done to make the previous suggestion less confusing :) but it seems to have had the opposite effect...

webtau-http/src/main/java/com/twosigma/webtau/http/HttpHeader.java
tsiq-karold: I get the comment on the type but for someone who isn't a webtau developer and is just looking at the docs, it'll be super confusing.

tsiq-karold:
@MykolaGolubyev curious if you're still planning on doing this?

MykolaGolubyev:
closing in favor of #1153

PR #467, opened 2019-12-10
add support for http patch

### LINKED ISSUE #466: support HTTP PATCH method
Our API would like to introduce PATCH methods and needs to test them using webtau.
We probably don't need the whole gamut of options like for put or post, but we definitely need to define query parameters and headers.

tsiq-clemens:
fixes #466

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: java
 
+jdk:
+  - openjdk8
+
 cache:
   directories:
   - $HOME/.m2
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,5 @@
 language: java
 
-jdk:
-  - openjdk8
-
 cache:
   directories:
   - $HOME/.m2
--- a/webtau-feature-testing/examples/scenarios/rest/springboot/customerCrud.groovy
+++ b/webtau-feature-testing/examples/scenarios/rest/springboot/customerCrud.groovy
@@ -19,7 +19,7 @@ scenario("CRUD operations for customer") {
     }
 
     http.get("/customers/${id}") {
-        firstName.should == firstName
+        firstName.should == "FN"
         lastName.should == changedLastName
     }
 
--- a/webtau-feature-testing/examples/scenarios/rest/springboot/customerCrud.groovy
+++ b/webtau-feature-testing/examples/scenarios/rest/springboot/customerCrud.groovy
@@ -19,11 +19,10 @@ scenario("CRUD operations for customer") {
     }
 
     def changedFirstName = "NFN"
-    http.patch("/customers/${id}", [firstName: changedFirstName]) {
-        firstName.should == changedFirstName // specifying body is optional
-    }
+    http.patch("/customers/${id}", [firstName: changedFirstName])
 
     http.get("/customers/${id}") {
+        firstName.should == changedFirstName
         lastName.should == changedLastName
     }
 
--- a/webtau-feature-testing/examples/scenarios/rest/springboot/customerCrud.groovy
+++ b/webtau-feature-testing/examples/scenarios/rest/springboot/customerCrud.groovy
@@ -18,6 +18,11 @@ scenario("CRUD operations for customer") {
         lastName.should == changedLastName // specifying body is optional
     }
 
+    def changedFirstName = "NFN"
+    http.patch("/customers/${id}", [firstName: changedFirstName]) {
+        firstName.should == changedFirstName // specifying body is optional
+    }
+
     http.get("/customers/${id}") {
         lastName.should == changedLastName
     }
--- a/webtau-feature-testing/examples/scenarios/rest/springboot/customerCrud.groovy
+++ b/webtau-feature-testing/examples/scenarios/rest/springboot/customerCrud.groovy
@@ -18,6 +18,11 @@ scenario("CRUD operations for customer") {
         lastName.should == changedLastName // specifying body is optional
     }
 
+    http.get("/customers/${id}") {
+        firstName.should == firstName
+        lastName.should == changedLastName
+    }
+
     def changedFirstName = "NFN"
     http.patch("/customers/${id}", [firstName: changedFirstName])
 
--- a/webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyOverloadsTest.groovy
+++ b/webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyOverloadsTest.groovy
@@ -91,6 +91,157 @@ class HttpGroovyOverloadsTest extends HttpTestBase {
     }
 
     @Test
+    void "patch with return overloads"() {
+        def v
+        v = http.patch("/full-echo", query, requestHeader, requestBodyMap) {
+            headerValidation.accept(body)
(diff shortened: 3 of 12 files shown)


tsiq-clemens:
Checked in functionality and a set of tests. Let me know if there's a set of tests you recommend I add, @MykolaGolubyev  and @tsiq-karold

webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
```diff
@@ -296,8 +315,9 @@ class HttpGroovyTest extends HttpTestBase {
             body.should == headerPayload
         }
 
-        http.get("/echo-header", requestHeader, expectations)
-        http.get("/echo-header", [qp1: 'v1'], requestHeader, expectations)
+//        http.get("/echo-header", requestHeader, expectations)
```
tsiq-karold: If these two aren't needed, can we just delete them rather than commenting out please?

webtau-http/src/main/java/com/twosigma/webtau/http/Http.java
```diff
@@ -730,6 +877,23 @@ private HttpResponse request(String method, String fullUrl,
         }
     }
 
+    private void setRequestMethod(String method, HttpURLConnection connection) throws ProtocolException {
+        if (method.equals("PATCH")) {
```
tsiq-karold: This is very sad :(

webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
tsiq-clemens: oops, my bad. I had trouble figuring out test failures and forgot to uncomment these lines. Back now.

webtau-http/src/main/java/com/twosigma/webtau/http/Http.java
tsiq-clemens: indeed. It gets slightly hackier if you need to support `patch` over https.
I guess this is Oracle's way to suggest moving up to the latest LTS (11)

tsiq-karold: Oh, so you're suggesting this approach won't work for PATCH https?  If that is the case I would suggest we fail loudly and explicitly, presumably right now we'll fail in weird and confusing ways.

tsiq-clemens: I think I can make it work for https, or otherwise make it scream in some obvious way.

tsiq-karold: If you need this quickly for http then let's scream but if it can wait for https support then I think we should do that.  Essentially, I'm saying up to you but I think we should do one or the other, the current approach will likely fail in really non-obvious ways (correct me if I'm wrong).

tsiq-clemens: We use https, so this needs to be working.

I just checked in the https work-around. I kept all the hacky stuff in a single method although it violates my personal preference for "deep nesting avoidance". Let me know what you think.

.travis.yml
```diff
@@ -1,5 +1,8 @@
 language: java
 
+jdk:
+  - openjdk8
```
tsiq-clemens: running with jdk 11 leads to a compile time error -- the https impl details are no longer accessible there.
Once webtau moves to jdk 11, we need to migrate to the recommended Builder solution.

webtau-feature-testing/examples/scenarios/rest/springboot/customerCrud.groovy
```diff
@@ -18,7 +18,11 @@ scenario("CRUD operations for customer") {
         lastName.should == changedLastName // specifying body is optional
     }
 
+    def changedFirstName = "NFN"
```
MykolaGolubyev: We used to call `get` right after `put` to make sure it took effect. Now we may hide the error in  `put`

webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyOverloadsTest.groovy
```diff
@@ -90,6 +90,157 @@ class HttpGroovyOverloadsTest extends HttpTestBase {
...
+
+        v = http.patch("/full-echo") {
+            pathValidation.accept(body)
+            return body[PATH_KEY]
+        }
+        v.should == PATH_EXPECTED_RETURN
+    }
+
+    @Test
+    void "patch without return overloads"() {
```
MykolaGolubyev: I am sorry and I thank you for the pain

webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
```diff
@@ -314,6 +334,11 @@ class HttpGroovyTest extends HttpTestBase {
             // assertions go here
         }
 
+        http.patch("/end-point", http.header('Accept', 'application/octet-stream'),
```
MykolaGolubyev: Just curios about `patch` specs, what is the intended behavior of patching with an octet stream?

webtau-http/src/main/java/com/twosigma/webtau/http/Http.java
```diff
@@ -56,14 +73,9 @@
...
-import static com.twosigma.webtau.WebTauCore.equal;
-import static com.twosigma.webtau.cfg.WebTauConfig.getCfg;
-import static com.twosigma.webtau.reporter.IntegrationTestsMessageBuilder.action;
-import static com.twosigma.webtau.reporter.IntegrationTestsMessageBuilder.urlValue;
-import static com.twosigma.webtau.reporter.TokenizedMessage.tokenizedMessage;
-import static java.nio.charset.StandardCharsets.UTF_8;
-import static java.util.stream.Collectors.toList;
+import javax.net.ssl.HttpsURLConnection;
+import org.apache.commons.io.IOUtils;
+import sun.net.www.protocol.https.HttpsURLConnectionImpl;
```
MykolaGolubyev: ouch :(

webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/HttpGroovyTest.groovy
tsiq-clemens: I guess it depends on what resource you are patching and how the server interprets/applies the patch.  The RFC (https://tools.ietf.org/html/rfc5789#section-2.2) is vague in that regard; it has this example:
```   
PATCH /file.txt HTTP/1.1
   Host: www.example.com
   Content-Type: application/example
   If-Match: "e0023aa4e"
   Content-Length: 100

   [description of changes]```

webtau-feature-testing/examples/scenarios/rest/springboot/customerCrud.groovy
tsiq-clemens: good point. Let me patch that test up ;-)

PR #481, opened 2020-01-01
Introduce http listeners

MykolaGolubyev:
```
public interface HttpListener {
    default void beforeFirstHttpCall() {
    }

    default void beforeHttpCall(String requestMethod,
                                String passedUrl,
                                String fullUrl,
                                HttpHeader requestHeader,
                                HttpRequestBody requestBody) {
    }

    default void afterHttpCall(String requestMethod,
                               String passedUrl,
                               String fullUrl,
                               HttpHeader requestHeader,
                               HttpRequestBody requestBody,
                               HttpResponse response) {
    }
}
```
`beforeFirstHttpCall` can be used to automatically bootstrap an embedded server

### DIFF
--- a/webtau-http/src/main/java/com/twosigma/webtau/http/Http.java
+++ b/webtau-http/src/main/java/com/twosigma/webtau/http/Http.java
@@ -37,6 +37,7 @@ import com.twosigma.webtau.http.datanode.DataNodeBuilder;
 import com.twosigma.webtau.http.datanode.DataNodeId;
 import com.twosigma.webtau.http.datanode.StructuredDataNode;
 import com.twosigma.webtau.http.json.JsonRequestBody;
+import com.twosigma.webtau.http.listener.HttpListeners;
 import com.twosigma.webtau.http.multipart.MultiPartFile;
 import com.twosigma.webtau.http.multipart.MultiPartFormData;
 import com.twosigma.webtau.http.multipart.MultiPartFormField;
@@ -651,7 +652,7 @@ public class Http {
 
         HttpValidationResult validationResult = new HttpValidationResult(requestMethod, url, fullUrl, fullHeader, requestBody);
 
-        TestStep<Void, R> step = createHttpStep(validationResult, requestMethod, fullUrl, httpCall, fullHeader, validator);
+        TestStep<Void, R> step = createHttpStep(validationResult, httpCall, validator);
         try {
             return step.execute(StepReportOptions.REPORT_ALL);
         } finally {
@@ -661,14 +662,23 @@ public class Http {
     }
 
     private <R> TestStep<Void, R> createHttpStep(HttpValidationResult validationResult,
-                                                 String requestMethod, String fullUrl, HttpCall httpCall,
-                                                 HttpHeader fullRequestHeader,
+                                                 HttpCall httpCall,
                                                  HttpResponseValidatorWithReturn validator) {
         Supplier<R> httpCallSupplier = () -> {
+            HttpResponse response = null;
             try {
                 long startTime = Time.currentTimeMillis();
-                HttpResponse response = httpCall.execute(fullUrl, fullRequestHeader);
-                response = followRedirects(requestMethod, httpCall, fullRequestHeader, response);
+
+                BeforeFirstHttpCallListenerTrigger.trigger();
+                HttpListeners.beforeHttpCall(validationResult.getRequestMethod(),
+                        validationResult.getUrl(), validationResult.getFullUrl(),
+                        validationResult.getRequestHeader(), validationResult.getRequestBody());
+
+                response = httpCall.execute(validationResult.getFullUrl(),
+                        validationResult.getRequestHeader());
+
+                response = followRedirects(validationResult.getRequestMethod(),
+                        httpCall, validationResult.getRequestHeader(), response);
 
                 long endTime = Time.currentTimeMillis();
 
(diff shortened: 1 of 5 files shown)


tsiq-karold:
@MykolaGolubyev what's the usecase you have in mind for the beforeFirst and for the before/after methods?  Shouldn't we also include a afterLast so that anything you setup in beforeFirst can be torn down?

webtau-http/src/main/java/com/twosigma/webtau/http/Http.java
```diff
@@ -964,4 +981,16 @@ private Object extractOriginalValue(Object v) {
     private interface HttpCall {
         HttpResponse execute(String fullUrl, HttpHeader fullHeader);
     }
+
+    private static class BeforeFirstHttpCallListenerTrigger {
+        static {
+            HttpListeners.beforeFirstHttpCall();
```
tsiq-karold: Why not put this in a static initialiser block for the `Http` class?  Alternatively, `Http` is essentially a singleton, we could make its constructor private and call this function in there.

webtau-http/src/main/java/com/twosigma/webtau/http/listener/HttpListener.java
```diff
@@ -0,0 +1,41 @@
...
+import com.twosigma.webtau.http.request.HttpRequestBody;
+
+public interface HttpListener {
+    default void beforeFirstHttpCall() {
+    }
+
+    default void beforeHttpCall(String requestMethod,
+                                String passedUrl,
+                                String fullUrl,
+                                HttpHeader requestHeader,
```
tsiq-karold: `HttpHeader` is still mutable, should we always pass in a copy to be safe?

https://github.com/twosigma/webtau/blob/master/webtau-http/src/main/java/com/twosigma/webtau/http/HttpHeader.java#L91

webtau-http/src/main/java/com/twosigma/webtau/http/Http.java
MykolaGolubyev: One can define`http.query` or `http.header` ahead of time and that will trigger a callback. Wanted to be on a more restricted side.

webtau-http/src/main/java/com/twosigma/webtau/http/listener/HttpListener.java
MykolaGolubyev: Good question. Since we deprecated the call, I think it is fare to leave the responsibility on the person calling the deprecated method?

MykolaGolubyev:
@tsiq-karold the intention is to let people to bootstrap an embedded server.  Additionally is to trace calls.
I thought about `afterTheLastCall`. Not sure how to reliably implement it except using `onshutdwon` kind of handler (similar to reports gen). On the other hand I assume that an embedded server will shutdown itself on exit anyway.

webtau-http/src/main/java/com/twosigma/webtau/http/Http.java
tsiq-karold: That would technically still be before the 1st http call 😄 but I see what you're getting at.

webtau-http/src/main/java/com/twosigma/webtau/http/listener/HttpListener.java
tsiq-karold: That is true.  Perhaps we should also carry out the promise from the deprecation comment and actually start throwing from that function.

tsiq-karold:
> @tsiq-karold the intention is to let people to bootstrap an embedded server. Additionally is to trace calls.

I get the need for the a `BeforeFirst`, I'm not sure I see what you would do in the `beforeHttpCall` though, what did you mean by trace?

> I thought about `afterTheLastCall`. Not sure how to reliably implement it except using `onshutdwon` kind of handler (similar to reports gen).

Similar to report gen could work.  Alternatively, we presumably know once we've executed all the scenarios so can't we just call it then?

> On the other hand I assume that an embedded server will shutdown itself on exit anyway.

In-process embedded server, yes.  However, it's possible that people would use the `BeforeFirst` to start an out of process server for example or populate a real deployment with some test data.  Such usecases would require a clean up step.

webtau-http/src/main/java/com/twosigma/webtau/http/Http.java
MykolaGolubyev: That may not be the case of JUnit like runners. It may load class, trigger its static inits including `http.header`, but then won't run any `http.get` like calls.

webtau-http/src/main/java/com/twosigma/webtau/http/listener/HttpListener.java
MykolaGolubyev: Good idea. I will add throw. Will we have to increase the major version?

MykolaGolubyev:
@tsiq-karold will expand on BeforeFirst a bit first. This is mainly for JUnit like runners. On a pure groovy side we tend to have more scripting around and it is more DevOps like. So it is less of an issue there to bootstrap. 

By tracing I meant simple user/framework additional logging. In a way similar to `StepReporter` but more focused on `http` and flat. 

Let me add `afterLast` with a bold comment that it could be way after the real last call. I will have to use `TestListener.afterAllTests` (which relies on shutdown handler for junit like runners)

MykolaGolubyev:
Hmm, `TestListener.afterAllTests` will only work when `@Webtau` is present on a test. Which is not good for his sort of listeners. `@Webtau` is optional in a sense, that if you don't have it, test class won't be part of a report. I don't have any good idea right now how to make it work for both Groovy and JUnit.

webtau-http/src/main/java/com/twosigma/webtau/http/listener/HttpListener.java
tsiq-karold: Technically but I'm not sure we really need to at this stage.  @tsiq-clemens do you use `HttpHeader.add` anywhere in your tests?

tsiq-karold:
OK, let's park the after last thing for now.  When someone asks for it, we can revisit.

webtau-http/src/main/java/com/twosigma/webtau/http/listener/HttpListener.java
tsiq-clemens: we merge headers at times, but I didn't see an `add` call

MykolaGolubyev:
@tsiq-karold good to go?

tsiq-karold:
> @tsiq-karold good to go?

Let's do it.

PR #487, opened 2020-01-07
add DocumentationArtifacts.createAsJson, createAsCsv

MykolaGolubyev:


webtau-core/src/main/java/com/twosigma/webtau/documentation/DocumentationArtifacts.java
```diff
@@ -16,13 +16,27 @@
...
 
 public class DocumentationArtifacts {
-    public static void create(Class testClass, String artifactName, String textContent) {
+    public static void create(Class<?> testClass, String artifactName, String text) {
         Path path = DocumentationArtifactsLocation.classBasedLocation(testClass).resolve(artifactName);
-        FileUtils.writeTextContent(path, textContent);
+        FileUtils.writeTextContent(path, text);
+    }
+
+    public static void createAsJson(Class<?> testClass, String artifactName, TableData tableData) {
```
tsiq-karold: Do we not want overrides for something other than table data?

MykolaGolubyev: Don't have usecases yet. There is a `String`  fallback

PR #489, opened 2020-01-08
add Record.get default value

MykolaGolubyev:


webtau-core/src/main/java/com/twosigma/webtau/data/table/header/Header.java
```diff
@@ -111,6 +111,16 @@ public String toString() {
...
     }
 
+    /**
+     * column index by column name
+     * @param columnName column name to get index for
+     * @return column index or -1 if not found
+     */
+    public int findColumnIdxByName(String columnName) {
+        Integer idx = indexByName.get(columnName);
+        return idx == null ? -1: idx;
```
tsiq-clemens: do we want to make `-1` a constant "COLUMN_NOT_FOUND" or some such?

webtau-core/src/main/java/com/twosigma/webtau/data/table/header/Header.java
```diff
@@ -111,6 +111,16 @@ public String toString() {
         return namesByIndex.toString();
     }
 
+    /**
+     * column index by column name
+     * @param columnName column name to get index for
+     * @return column index or -1 if not found
+     */
+    public int findColumnIdxByName(String columnName) {
+        Integer idx = indexByName.get(columnName);
```
tsiq-clemens: out of curiosity: what does indexByName.get(null) return? null or exception?

webtau-core/src/test/groovy/com/twosigma/webtau/data/table/RecordTest.groovy
```diff
@@ -39,4 +39,24 @@ class RecordTest {
         def record = new Record(new Header(["c1", "c2", "c3"].stream()), ["v1", "v2", "v3"].stream())
         assert record.key == null
     }
+
+    @Test
+    void "should return defaultValue if a columnName is not present"() {
+        def record = new Record(new Header(["c1", "c2", "c3"].stream()), ["v1", "v2", "v3"].stream())
+        def valueC1 = record.get("c1", 42)
+        def valueC4 = record.get("c4", 42)
```
tsiq-clemens: or `get("noSuchColumn", 42)`?

webtau-core/src/test/groovy/com/twosigma/webtau/data/table/RecordTest.groovy
```diff
@@ -39,4 +39,24 @@ class RecordTest {
...
+        assert valueC1 == "v1"
+        assert valueC4 == 42
+    }
+
+    @Test
+    void "should return defaultValue if an idx is not present"() {
+        def record = new Record(new Header(["c1", "c2", "c3"].stream()), ["v1", "v2", "v3"].stream())
+        def valueC2 = record.get(1, 42)
+        def valueCN = record.get(5, 42)
```
tsiq-clemens: why 5? 3 is the edge case. I'd also indicate the too-high-index in the variable name

webtau-core/src/main/java/com/twosigma/webtau/data/table/header/Header.java
MykolaGolubyev: Not to keen on that. I think by now -1 is a common knowledge.

webtau-core/src/test/groovy/com/twosigma/webtau/data/table/RecordTest.groovy
MykolaGolubyev: Intention here is to show more like columns that can be there but not in this case. I.e optional table columns. But I see what you mean. Sometimes I do that.

webtau-core/src/test/groovy/com/twosigma/webtau/data/table/RecordTest.groovy
MykolaGolubyev: 5 cause more than 2 :)
I feel like the scope of this test is to small to introduce named vars. If you feel strongly about it, I will change.

webtau-core/src/main/java/com/twosigma/webtau/data/table/header/Header.java
MykolaGolubyev: null

webtau-core/src/test/groovy/com/twosigma/webtau/data/table/RecordTest.groovy
MykolaGolubyev: Ok, changed :)

Issue #488, opened 2020-01-07
Test metadata annotation


tsiq-brian:
Our integration tests are authored and maintained by different teams. Tests are currently organized by functionality, and there is no simple functionality->team association.

We'd like some way of injecting static metadata into the tests so that we can use it in reporting and somehow ensure that all tests have an owner (or fail somewhere if not). The use cases we have are...

* propagating test ownership data (a string, array of strings, or something) to the report json that gets generated so that we can notify those specific teams when their integration test fails.
* enforcing each test has ownership data associated so that no one can push a test that isn't owned by someone

For the JUnit Java/Groovy route, we can create a custom annotation and annotate the test classes/methods. For the two use cases, we can probably scan the classpath for those annotations and have a custom reporter and enforcer unit test pull that metadata.

However, most of our tests use the Groovy DSL. Maybe it's my lack of Groovy knowledge, but I don't know if there's a straightforward way to "annotate" the `scenario` call in the test, or have that data available within other tests.

tsiq-clemens:
The report json is custom-generated, so we need
* some metadata object attachable to a `WebTauTest`
* a way to retrieve that object from the `WebTauTest` so that custom generators can do something with it

tsiq-clemens:
For the "fail somewhere if not part", the idea is that either 
* a custom webtau test fails if not all tests have an owner (but what about this custom test's owner?) 
* the webtau suite as a whole fails
* some process that runs the webtau suite fails

MykolaGolubyev:
Interesting request! Thank you.
You can define custom values in config. Config values should be part of a report. Owner custom config value could be it.

There are also multiple listeners one can attach custom code to. You can have custom httpHeaderValues injected into each call. You can listen for each http call and react on it.

So if each team has its own webtau config file then add a new `owner` value there.
As part of http listeners you can listen to a call, and exit hard if `cfg.owner` is not defined. 
Registering http calls listener is only available right now through `META-INF/services`. Potentially you could “attach” a jar file with impl during test run on your side to enforce the rule.  Or you can provide a company wide wrapper to run `webtau` that enforces the rule.

In custom generated report you can access `owner` value using `cfg`.

tsiq-clemens:
More background:
We have a single "api integration tests" suite with a single webtau.cfg file, maintained by multiple teams.
The multiple-config-files approach won't work in that case.

tsiq-karold:
[...]

PR #491, opened 2020-01-11
Groovy test runner test metadata

MykolaGolubyev:
Addresses groovy part of #488 
image
image


webtau-core/src/main/java/com/twosigma/webtau/reporter/WebTauTestMeta.java
```diff
@@ -0,0 +1,50 @@
...
+    public void add(WebTauTestMeta meta) {
+        this.meta.putAll(meta.meta);
+    }
+
+    public boolean isEmpty() {
+        return meta.isEmpty();
+    }
+
+    public Map<String, Object> toMap() {
+        return meta;
```
tsiq-karold: This probably should be a copy or an immutable wrapper of the map.

webtau-feature-testing/examples/scenarios/concept/metaDataMethodBased.groovy
```diff
@@ -0,0 +1,15 @@
+package scenarios.concept
+
+import static com.twosigma.webtau.WebTauGroovyDsl.*
+
+Support.teamA()
```
tsiq-karold: I think I'd still be quite interested in syntax that looks more similar to:
```
metadata {
  team = "foo",
  owner = "blah",
...
}
```
Is that possible with the current API?  Main reason why I like this syntax is it allows you to group and scope all metadata specification together.  Otherwise you'll end up with lots of:
```
Support.teamA()
Support.foo()
Support.blah()
```
This gets a little tedious and worse, someone might end up scattering it all over the file.

tsiq-karold: Hmm, though I suppose that makes the scenario level overriding a little more difficult.

webtau-feature-testing/src/main/groovy/com/twosigma/webtau/featuretesting/WebTauEndToEndTestRunner.groovy
```diff
@@ -129,9 +129,15 @@ class WebTauEndToEndTestRunner implements StepReporter, TestListener {
...
     @Override
     void afterTestRun(WebTauTest test) {
-        scenariosDetails.add([scenario        : test.scenario,
-                              shortContainerId: test.shortContainerId,
-                              stepsSummary    : capturedStepsSummary])
+        def details = [scenario        : test.scenario,
+                       shortContainerId: test.shortContainerId,
+                       stepsSummary    : capturedStepsSummary]
+
+        if (!test.meta.isEmpty()) {
```
tsiq-karold: Why not just add an empty map to details?

webtau-docs/webtau/groovy-specific-runner/test-meta.md
```diff
@@ -0,0 +1,21 @@
+# Specifying Test Meta
```
tsiq-clemens: Metadata

webtau-core/src/main/java/com/twosigma/webtau/reporter/WebTauTestMeta.java
MykolaGolubyev: Ah yes. It was and I reversed it :( will change back.

webtau-feature-testing/examples/scenarios/concept/metaDataMethodBased.groovy
MykolaGolubyev: It is possible to have different syntaxes. As in the end it boils down to calling a method one can come up with various shortcuts or DSLs. 
If we feel like we need to support metadata closure think that it would be better if we support meta closure out of the box I will add it to the available methods. 

In case you described above test writers can opt for a shortcut like ‘Support.meta(k: v, a: 10)’. 

I wanted to have a syntax that can support strongly defined keys and values.

webtau-feature-testing/src/main/groovy/com/twosigma/webtau/featuretesting/WebTauEndToEndTestRunner.groovy
MykolaGolubyev: To avoid multiple broken tests and base line updates. Since meta is not required part of a scenario I figured it will be more convenient to avoid adding it if it is not preset.

webtau-docs/webtau/groovy-specific-runner/test-meta.md
```diff
@@ -0,0 +1,21 @@
...
+
+:include-file: scenarios/concept/Support.groovy { title: "Support.groovy"}
+
+# Report
+
+Test meta data will be available in a generated report as well as in test listeners.
+
+:include-image: doc-artifacts/reports/report-test-meta.png {annotationsPath: "doc-artifacts/reports/report-test-meta.json"}
+
+1. If meta is present, it will be displayed at the test summary page
```
tsiq-clemens: If meta data is present, ...

webtau-feature-testing/examples/scenarios/concept/metaDataMethodBased.groovy
tsiq-karold: Curious to hear @tsiq-clemens' thoughts on this too as he's the one who actually has a use case for this :)

webtau-feature-testing/src/main/groovy/com/twosigma/webtau/featuretesting/WebTauEndToEndTestRunner.groovy
tsiq-karold: Ah, yes, I remember having this problem when I changed something about the test output capture json.  Maybe we should revisit how that works at some point but that's for a different day/PR.

webtau-docs/webtau/groovy-specific-runner/test-meta.md
tsiq-karold: In general, I think I prefer metadata over meta data but happy to be convinced otherwise.

webtau-feature-testing/examples/scenarios/concept/metaDataMethodBased.groovy
tsiq-clemens: I can see scenario-level overrides in the shape of an optional 2nd parameter (map), e.g.
`scenario('some scenario name', [ owner: 'scenario owner' ]) { ... }`
The implementation would effectively walk through the map calling `attachTestMetaValue(key, 

I am not sure that that would look consistent with the method-syntax to the end user / developer, though.

webtau-feature-testing/src/main/groovy/com/twosigma/webtau/featuretesting/WebTauEndToEndTestRunner.groovy
tsiq-clemens: +1

webtau-docs/webtau/groovy-specific-runner/test-meta.md
tsiq-clemens: let's go with Merrian-Webster and make it a single word, no space or hyphen: https://www.merriam-webster.com/dictionary/metadata

webtau-feature-testing/examples/scenarios/concept/metaDataMethodBased.groovy
tsiq-clemens: I quite like the declarative nature of the closure syntax; if it's easy to support it has my vote. 
If it's a lot of work, you can get a bit more declarative with the method calls as well, e.g. `Support.supportedByTeam("Billing")` or even more strongly defined `Support.supportedByBilling()`.
There is a danger of scattering those function calls, of course; similarly, you could have multiple metadata closures and mess up the code. At some point, we need to appeal to the Clean Coder in the test developer to keep things organized.

tsiq-karold: With the closure you at least have the option of throwing on the second invocation

MykolaGolubyev:
I don’t mind either implementation and they both can share the already existing core to implement a different syntax. 
We settled on a syntax in the relevant issue and that’s why I went with it. 

Maybe we should get back to the drawing board and discuss it there.

webtau-feature-testing/examples/scenarios/concept/metaDataMethodBased.groovy
MykolaGolubyev: Can you elaborate on second invocation? Let’s move it back to the issue where we discussed the syntax. I may not fully understood the problem.

webtau-core/src/main/java/com/twosigma/webtau/reporter/WebTauTestMetadata.java
```diff
@@ -17,27 +17,35 @@
 package com.twosigma.webtau.reporter;
 
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
-public class WebTauTestMeta {
+public class WebTauTestMetadata {
```
tsiq-clemens: +1  This new name makes more sense

webtau-core/src/main/java/com/twosigma/webtau/reporter/WebTauTestMetadata.java
```diff
@@ -17,27 +17,35 @@
...
         meta = new LinkedHashMap<>();
     }
 
     /**
      * register key/value meta
-     * @param key key
-     * @param value value
-     * @return previously registered value for key
+     * @param values metadata values to add
+     * @return previously registered values
```
tsiq-clemens: remind me what we need the previous values for?

webtau-docs/webtau/groovy-specific-runner/test-metadata.md
```diff
@@ -0,0 +1,21 @@
+# Specifying Test Metadata
+
+To attach custom meta data to a test use `metadata` from `WebTauGroovyDsl`
```
tsiq-clemens: typo: metadata

webtau-docs/webtau/groovy-specific-runner/test-metadata.md
```diff
@@ -0,0 +1,21 @@
...
+ 
+:include-file: scenarios/concept/metadataRaw.groovy { title: "attach test metadata" }
+
+# Meta Data Encapsulation
```
tsiq-clemens: typo: Metadata

webtau-docs/webtau/groovy-specific-runner/test-metadata.md
```diff
@@ -0,0 +1,21 @@
...
+
+You can encapsulate repeated metadata into a separate file. 
+ 
+:include-file: scenarios/concept/metadataMethodBased.groovy { title: "encapsulated metadata" }
+
+:include-file: scenarios/concept/Support.groovy { title: "Support.groovy"}
+
+# Report
+
+Test meta data will be available in a generated report as well as in test listeners.
```
tsiq-clemens: typo: metadata

webtau-feature-testing/test-expectations/scenarios/concept/metadataMethodBased/run-details.json
```diff
@@ -0,0 +1,25 @@
+{
+  "scenarioDetails" : [ {
+    "scenario" : "one",
+    "shortContainerId" : "metadataMethodBased.groovy",
+    "stepsSummary" : { },
+    "meta" : {
```
tsiq-clemens: should be "metadata", no?

webtau-reactjs/src/report/details/TestSummary.demo.js
```diff
@@ -0,0 +1,55 @@
...
+        "elapsedTime": 123,
+        "fileName": "scenarios/cli/simpleRun.groovy",
+        "disabled": false,
+    }
+}
+
+function withMeta() {
+    return {
+        ...sampleTest(),
+        meta: {
```
tsiq-clemens: metadata

webtau-core/src/main/java/com/twosigma/webtau/reporter/WebTauTestMetadata.java
MykolaGolubyev: To potentially fail hard in case of custom metadata wrappers.

MykolaGolubyev:
image

webtau-docs/webtau/groovy-specific-runner/test-metadata.md
```diff
@@ -0,0 +1,27 @@
+# Specifying Test Metadata
+
+To attach custom metadata to a test use `metadata` from `WebTauGroovyDsl`
+ 
+:include-file: scenarios/concept/metadataRaw.groovy { title: "test metadata", commentsType: "inline" }
+
+`metadata` can be called multiple times outside of `scenario`. It will be applied to all the scenarios below.
```
tsiq-clemens: by "below" I assume you mean "after the invocation"?

tsiq-clemens: or better "following the invocation"

webtau-groovy/src/main/groovy/com/twosigma/webtau/WebTauGroovyDsl.groovy
```diff
@@ -75,6 +75,16 @@ class WebTauGroovyDsl extends WebTauDsl {
         }
     }
 
+    static Map<String, Object> metadata(Map<String, Object> meta) {
+        if (testRunner) {
+            return testRunner.attachTestMetadata(meta)
+        } else {
+            ConsoleOutputs.out(Color.YELLOW, "ignoring setting meta in ad-hoc mode ", Color.PURPLE,
```
tsiq-clemens: metadata

PR #590, opened 2020-07-26
doc capture: validate that artifact name is unique

### LINKED ISSUE #495: Protect against reuse of artifact name in doc.capture
If multiple calls to `http.doc.capture` (or other capture apis) use the same name, they'll end up stepping over each other.  We should throw if a name is reused.

MykolaGolubyev:
closed #495

### DIFF
--- a/webtau-core/src/main/java/org/testingisdocumenting/webtau/documentation/DocumentationArtifacts.java
+++ b/webtau-core/src/main/java/org/testingisdocumenting/webtau/documentation/DocumentationArtifacts.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright 2020 webtau maintainers
  * Copyright 2019 TWO SIGMA OPEN SOURCE, LLC
  *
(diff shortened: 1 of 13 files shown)


webtau-core/src/main/java/org/testingisdocumenting/webtau/documentation/DocumentationArtifacts.java
```diff
@@ -23,8 +24,18 @@
 import org.testingisdocumenting.webtau.utils.JsonUtils;
 
 import java.nio.file.Path;
+import java.util.concurrent.ConcurrentHashMap;
 
 public class DocumentationArtifacts {
+    private static final ConcurrentHashMap<String, Boolean> usedArtifactNames = new ConcurrentHashMap<>();
+
+    public static void registerName(String artifactName) {
+        Boolean previous = usedArtifactNames.put(artifactName, true);
```
tsiq-karold, line 33: I think a set of names with a contains check before a put would be a little clearer but probably not worth changing now.

PR #599, opened 2020-08-05
Initial graphql dsl support

### LINKED ISSUE #585: GraphQL DSL - initial support
* check for 200 instead of 201 as implied by requests being POSTs
* better response data access
* no need to always specify url
* 1st class knowledge of query, variables, etc.

tsiq-karold:
I know I haven't added any tests, I will do that before I merge this PR.  However, I wanted to open it now to start a discussion about the DSL so comments very welcome 😁 

Some questions, comments, food for thought, etc. below.  Once we have some sort of consensus I can break these up into issues.

### Handling of `errors`

I'm in two minds about how to handle the `errors`.  Initially I was going to have default assertions if the user does not provide any (like status code for example).

However, that may be annoying because there may be cases where you just don't care whether there are errors if all else is good so you'd want to disable the default handling.  I'm not sure that's a common case though so perhaps the default handling makes sense.

### Query construction helpers

I think it would make sense to offer a better way to provide the query rather than as a string.  Perhaps we can construct it from a map or some other custom DSL support.

### CLI rendering of requests and responses

I'd like to override the rendering of queries and potentially responses.  Right now it'll print the json request as follows:

```
> executing HTTP POST http://127.0.0.1:52641/graphql
request (application/json):
{
  "query": "
query {
    allTasks(uncompletedOnly: false) {
        id
        description
    }
}
"
}
```

In this example it's just about ok but that's only because I formatted the input string.  I think we should override that but it'll require some changes to http handling to allow overriding the rendering.

Response rendering is less of an issue as responses are basically pure json but may make sense to have custom rendering for data and errors rather than the raw json.

### CLI rendering of assertions

Validations show up as follows:
```
  . body.data.allTasks.id equals [a, b, c]
      matches:
      
      body.data.allTasks.id[0]:   actual: "a" <java.lang.String>
                                expected: "a" <java.lang.String>
      body.data.allTasks.id[1]:   actual: "b" <java.lang.String>
                                expected: "b" <java.lang.String>
      body.data.allTasks.id[2]:   actual: "c" <java.lang.String>
                                expected: "c" <java.lang.String> (7ms)

```

My test makes no mention of `body` or `data` so this *could* be a little confusing.  That said, I think it's the same for pure HTTP where `foo.should == 1` shows up as `body.foo` so I'm ok with this.

### Implicit status code check

~The implicit status code check is printed to the screen which is not the case for the default HTTP status code checks.  We may want a different mechanism for performing this check.~ Turns out this is the same behaviour as with pure HTTP so will leave it as is.

### Missing features

1. much overriding for various combinations of query, variables, op name, headers, etc., sigh
1. some unit tests 😁 
1. some configurability though we may wish to wait until people ask for these things:
    1. url - in theory it's possible someone may serve at root for example, rather than `graphql`.  It's debatable whether we should do this at all or just force users to stick the `graphql` path in the base url config.
    1. success status code - in case people deviate from best practices
    1. invocation method - there are many ways to invoke a graphql service according to best practices: POST bodies like I've used, GET and query params and POST with `application/graphql` content type.  We probably should allow any of these.

Fixes #585

### DIFF
--- a/webtau-feature-testing/examples/scenarios/graphql/queryAndMutation.groovy
+++ b/webtau-feature-testing/examples/scenarios/graphql/queryAndMutation.groovy
@@ -28,36 +28,21 @@ mutation complete($id: ID!) {
 '''
 
 scenario("list all tasks") {
-    http.post("/graphql", [query: listAllQuery]) {
-        statusCode.should == 200
+    graphql.execute(listAllQuery) {
         errors.should == null
-        body.data.allTasks.id.should == ["a", "b", "c"]
+        allTasks.id.should == ["a", "b", "c"]
     }
 }
 
 scenario("complete a task") {
-    def completionRequest = [
-        query: completeMutation,
-        variables: [
-            id: "a"
-        ]
-    ]
-    http.post("/graphql", completionRequest) {
-        statusCode.should == 200
+    graphql.execute(completeMutation, [id: "a"]) {
         errors.should == null
-        body.data.complete.should == true
+        complete.should == true
     }
 
-    def getTaskRequest = [
-        query: taskByIdQuery,
-        variables: [
-            id: "a"
-        ]
-    ]
-    http.post("/graphql", getTaskRequest) {
-        statusCode.should == 200
+    graphql.execute(taskByIdQuery, [id: "a"]) {
         errors.should == null
-        body.data.taskById.id.should == "a"
-        body.data.taskById.completed.should == true
+        taskById.id.should == "a"
+        taskById.completed.should == true
     }
 }
--- a/webtau-feature-testing/examples/scenarios/graphql/queryAndMutation.groovy
+++ b/webtau-feature-testing/examples/scenarios/graphql/queryAndMutation.groovy
@@ -47,6 +47,7 @@ scenario("list all tasks") {
     graphql.execute(listAllQuery) {
         errors.should == null
         allTasks.id.should == ["a", "b", "c"]
+        body.data.allTasks.id.should == ["a", "b", "c"]
     }
 }
 
--- a/pom.xml
+++ b/pom.xml
@@ -135,6 +135,7 @@
         <module>webtau-dist</module>
         <module>webtau-cli-testing</module>
         <module>webtau-graphql</module>
+        <module>webtau-graphql-groovy</module>
     </modules>
 
     <properties>
--- a/webtau-feature-testing/pom.xml
+++ b/webtau-feature-testing/pom.xml
@@ -59,6 +59,12 @@
 
         <dependency>
             <groupId>org.testingisdocumenting.webtau</groupId>
+            <artifactId>webtau-graphql-groovy</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.testingisdocumenting.webtau</groupId>
             <artifactId>webtau-utils</artifactId>
             <version>${project.version}</version>
         </dependency>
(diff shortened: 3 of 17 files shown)


MykolaGolubyev:
Do we have an example of this DSL in Java? Also forgot to ask if there maybe a reason to have a separate method for mutation vs query.

tsiq-karold:
> Do we have an example of this DSL in Java?

Not yet, I was going to add equivalents of `HttpJavaTest` and `HttpGroovyTest` before merging this PR.



> Also forgot to ask if there maybe a reason to have a separate method for mutation vs query.

Yeah, I considered this and decided not to just yet.  I think ultimately we probably will once we build some support/utils for building up a query.

tsiq-karold:
As you've probably seen, I converted a bunch of the things we discussed into issues in https://github.com/testingisdocumenting/webtau/projects/4.

webtau-feature-testing/examples/scenarios/graphql/queryAndMutation.groovy
```diff
@@ -44,36 +44,21 @@ mutation complete($id: ID!) {
 '''
 
 scenario("list all tasks") {
-    http.post("/graphql", [query: listAllQuery]) {
-        statusCode.should == 200
+    graphql.execute(listAllQuery) {
         errors.should == null
-        body.data.allTasks.id.should == ["a", "b", "c"]
+        allTasks.id.should == ["a", "b", "c"]
```
tsiq-clemens, line 49: makes sense to me to allow skipping the `body.data` portion, that is similar to how you can omit the `body` for http validations. 
That said, shall we also allow the "full path", i.e., `body.data.allTasks.id` ?

webtau-feature-testing/examples/scenarios/graphql/queryAndMutation.groovy
```diff
@@ -44,36 +44,21 @@ mutation complete($id: ID!) {
...
 scenario("complete a task") {
-    def completionRequest = [
-        query: completeMutation,
-        variables: [
-            id: "a"
-        ]
-    ]
-    http.post("/graphql", completionRequest) {
-        statusCode.should == 200
+    graphql.execute(completeMutation, [id: "a"]) {
```
tsiq-clemens, line 54: this syntax (query/mutation, optional variables) makes sense to me

tsiq-clemens:
> Also forgot to ask if there maybe a reason to have a separate method for mutation vs query.
It does make sense to have, similar to how we do `http.get`, `http.post` etc.   
Because they're part of the query string, they could only be meaningfully used in combination with constructing the query/mutation string from some data structure.

webtau-feature-testing/examples/scenarios/graphql/queryAndMutation.groovy
tsiq-karold: Good point, yes, I think we should.

tsiq-karold:
> > Do we have an example of this DSL in Java?
> 
> Not yet, I was going to add equivalents of `HttpJavaTest` and `HttpGroovyTest` before merging this PR.

I've added these now.  There are some rather annoying things in the Java version though:
* calls to `body.get()` have to specify `data`
* query string construction is painful due to lack of multi-line strings
* the need to specify the query name in the assertion path is even more annoying than in Groovy as the whole expression is already longer.  See https://github.com/testingisdocumenting/webtau/issues/606.

Also, I realise that there is a LOT of overlap between `GraphQLTestDataServer` and `WebTauGraphQLFeaturesTestData`.  I propose to leave it as is for now and address in a future PR.

tsiq-karold:
LOL, I've just noticed that virtually all the new code I've put in place for GraphQL has gone into the `org.testgisdocumenting.webtau.graphql` package.  Guess none of us noticed the typo 😁 

I'll fix that in a separate PR once I've merged this one.  Otherwise this PR will become unreadable.

PR #612, opened 2020-08-10
Fix GraphQL query parsing

tsiq-karold:
This changes two things:
1. fixes local build issues where `mvn package/install` would fail.  This was failing because `GraphQLReportDataProvider` grabbed a reference to `GraphQLCoverage` from `GraphQL` when the data provider was first constructed.  However, during tests, `GraphQL.reset()` ends up being invoked a number of times but nothing recreates the data provider.  Therefore we were recording the correct coverage, just not on the coverage object that the report data provider was using.
1. I had misunderstood slightly the internal representation of queries within the library we use for parsing.  I've addressed this and added some appropriate tests.  Unfortunately, we have overloaded the word "operation' and webtau's meaning of it (the name/type tuple, e.g. `query foo`) differs from GraphQL's meaning (which is essentially a collection of queries, mutations or subscriptions but not mixed).  Our choice of operation here was for symmetry with OpenApi but if someone has a better suggestion for naming to remove the confusion then I'm all ears.

### DIFF
--- a/webtau-feature-testing/examples/scenarios/graphql/queryAndMutation.groovy
+++ b/webtau-feature-testing/examples/scenarios/graphql/queryAndMutation.groovy
@@ -18,8 +18,16 @@ package scenarios.graphql
 
 import static org.testingisdocumenting.webtau.WebTauGroovyDsl.*
 
-def listAllQuery = '''
+/*def listAllQuery = '''
 query {
+    allTasks(uncompletedOnly: false) {
+        id
+        description
+    }
+}
+'''*/
+def listAllQuery = '''
+query allTasks {
     allTasks(uncompletedOnly: false) {
         id
         description
--- a/webtau-feature-testing/examples/scenarios/graphql/queryAndMutation.groovy
+++ b/webtau-feature-testing/examples/scenarios/graphql/queryAndMutation.groovy
@@ -18,16 +18,8 @@ package scenarios.graphql
 
 import static org.testingisdocumenting.webtau.WebTauGroovyDsl.*
 
-/*def listAllQuery = '''
-query {
-    allTasks(uncompletedOnly: false) {
-        id
-        description
-    }
-}
-'''*/
 def listAllQuery = '''
-query allTasks {
+{
     allTasks(uncompletedOnly: false) {
         id
         description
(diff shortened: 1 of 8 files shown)


webtau-feature-testing/examples/scenarios/graphql/queryAndMutation.groovy
```diff
@@ -19,7 +19,7 @@ package scenarios.graphql
 import static org.testingisdocumenting.webtau.WebTauGroovyDsl.*
 
 def listAllQuery = '''
-query {
+{
```
tsiq-karold, line 22: I didn't actually need to do this, was left over from debugging, but it's correct either way.

webtau-graphql/src/main/java/org/testingisdocumenting/webtau/graphql/GraphQLReportDataProvider.java
```diff
@@ -28,27 +28,28 @@
...
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public class GraphQLReportDataProvider implements ReportDataProvider {
-    private final GraphQLCoverage coverage;
+    private final Supplier<GraphQLCoverage> coverage;
 
     public GraphQLReportDataProvider() {
-        this(GraphQL.getCoverage());
+        this(GraphQL::getCoverage);
```
tsiq-karold, line 39: This is what fixed the local mvn run problem.  I can only assume in Travis we somehow fork a jvm per test class.  It's the only explanation I have for this problem not happening there.

webtau-graphql/src/main/java/org/testingisdocumenting/webtau/graphql/GraphQLReportDataProvider.java
```diff
@@ -28,27 +28,28 @@
 import java.util.LongSummaryStatistics;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public class GraphQLReportDataProvider implements ReportDataProvider {
-    private final GraphQLCoverage coverage;
+    private final Supplier<GraphQLCoverage> coverage;
```
MykolaGolubyev, line 36: shall we rename coverage to coverageSupplier or something like that?

PR #615, opened 2020-08-11
Rename GraphQL operation to query

tsiq-karold:
The word operation in GraphQL means something slightly different to what we had here.  The closest word to the <name/type> tuple is query.  That's slightly confusing in that a query has a type which could be query but could be something else but it's in line with GraphQL best practices (e.g. the field in the HTTP request body is called query).


webtau-graphql/src/test/groovy/org/testingisdocumenting/webtau/graphql/GraphQLCoverageTest.groovy
```diff
@@ -34,13 +34,13 @@ class GraphQLCoverageTest {
...
-        coverage.nonCoveredOperations().should == ['*name'     | '*type'    ] {
-                                                   ___________________________________________
-                                                    'taskById' | GraphQLOperationType.QUERY
-                                                  'uncomplete' | GraphQLOperationType.MUTATION }
+    void "should provide non covered queries"() {
+        coverage.recordQuery(validationResult('allTasks', GraphQLQueryType.QUERY))
+        coverage.recordQuery(validationResult('complete', GraphQLQueryType.MUTATION))
+
+        coverage.nonCoveredQueries().should == ['*name'     | '*type'    ] {
+                                                   ____________________________________
```
MykolaGolubyev, line 42: shifted underscore

webtau-graphql/src/test/groovy/org/testingisdocumenting/webtau/graphql/GraphQLSchemaTest.groovy
```diff
@@ -118,9 +118,9 @@ query hello {
         def payload = [
             query: query
         ]
-        Set<GraphQLOperation> operations = schema.findOperations(new JsonRequestBody(payload))
+        Set<GraphQLQuery> queries = schema.findQueries(new JsonRequestBody(payload))
         // Multiple named operations without specifying an operation name in request are not valid
```
MykolaGolubyev, line 122: is comment still valid?

webtau-graphql/src/test/groovy/org/testingisdocumenting/webtau/graphql/GraphQLCoverageTest.groovy
tsiq-karold: Man, the table formatting always gets me.  We need to find a way to automate it as it's quite painful to maintain by hand.

webtau-graphql/src/test/groovy/org/testingisdocumenting/webtau/graphql/GraphQLSchemaTest.groovy
tsiq-karold: Yes.

webtau-graphql/src/test/groovy/org/testingisdocumenting/webtau/graphql/GraphQLCoverageTest.groovy
MykolaGolubyev: Totally agree. The best thing I know is custom IDEA formatter via plugin. Another best thing is to make webtau more popular and ask JetBrains to help.

tsiq-karold: The first option is definitely faster :)

PR #616, opened 2020-08-11
Load GraphQL schema via introspection

tsiq-karold:
There are a few issues with using a schema file:
1. not everyone will have one, it's often generated at runtime
1. splitting schema across files is common and merging at runtime
1. file has much more info than we need

In this PR I'm changing the gathering of declared queries to using GraphQL introspection to just give me a list of query, mutation and subscription names.

There were a couple of challenges with this:
1. when GraphQL stuff in webtau is initialised, webtau has not yet processed CLI args, only cfg file.  Therefore the url may be wrong.  I worked around this by making the loading lazy.
1. calling http from GraphQL initialisation caused an infinite loop.  GraphQL schema called http, this resulted in validators being invoked which tried to use the GraphQL schema which tried to call http, etc.  I worked around this by disabling the GraphQL HTTP validator while loading the schema.


webtau-graphql/src/main/java/org/testingisdocumenting/webtau/graphql/model/Request.java
```diff
@@ -0,0 +1,47 @@
...
+import org.testingisdocumenting.webtau.http.json.JsonRequestBody;
+import org.testingisdocumenting.webtau.http.request.HttpRequestBody;
+import org.testingisdocumenting.webtau.utils.StringUtils;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.testingisdocumenting.webtau.utils.CollectionUtils.notNullOrEmpty;
+
+public class Request {
```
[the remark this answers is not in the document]
tsiq-karold: Makes sense.

webtau-utils/src/main/java/org/testingisdocumenting/webtau/utils/JsonUtils.java
```diff
@@ -129,4 +129,8 @@ public static Object deserialize(String json) {
     public static JsonNode convertToTree(Object object) {
         return mapper.valueToTree(object);
     }
+
+    public static <T> T convert(Object fromValue, Class<T> toValueType) {
```
MykolaGolubyev, line 133: do we need a test for this?

tsiq-karold: We could add one but given it's an incredibly straightforward passthrough I'm inclined not to.

PR #621, opened 2020-08-12
graphql - more robust query processing

tsiq-karold:
Filter out more things which aren't GraphQL queries from coverage recording.


tsiq-karold:
@MykolaGolubyev this is independent of the problem I demonstrated in https://github.com/testingisdocumenting/webtau/compare/graphql-with-http-header-provider

PR #622, opened 2020-08-12
graphql: fix bug with schema loading in the presence of an http header provider

tsiq-karold:
What was happening is as this:
* scenario makes an http or graphql call
* webtau invokes the header provider, it does this within `HttpConfigurations.withDisabledConfigurations`
* header provider makes an http call
* webtau process response from the header provider http call and invokes all `HttpValidationHandler`s
* `GraphQLCoverageRecorder` (an impl of `HttpValidationHandler`) tries to query the GraphQL schema
* lazy loading of GraphQL schema is triggered
* graphql introspection call is made which utilises `HttpConfigurations.fullUrl` to prepend the configured url
* this fails because configurations are disabled so it tries to `http POST /graphql`

I've fixed this by adding a method to `HttpConfigurations` which is identical to `fullUrl` but without the enabled check.  I think this is fine for now but we should later revisit this.  I understand the reason for the enabled check in `fullHeader` to avoid the circular loop with the header provider but I'm not sure it's needed for the url construction.


webtau-http/src/main/java/org/testingisdocumenting/webtau/http/config/HttpConfigurations.java
```diff
@@ -47,10 +47,14 @@ public static void remove(HttpConfiguration configuration) {
...
     public static String fullUrl(String url) {
-        if (! enabled.get()) {
+        if (!enabled.get()) {
             return url;
         }
 
+        return fullUrlWithoutEnabledCheck(url);
+    }
+
+    public static String fullUrlWithoutEnabledCheck(String url) {
```
tsiq-clemens, line 57: might be worth adding a few comments why you'd want to run fullUrl vs fullUrlWithoutEnabledCheck

MykolaGolubyev: next step would be to separate those two and hopefully there won't be a need. @tsiq-karold did you by any chance created a ticket for this?

tsiq-karold: I haven't.  By those two you mean url and header, right?

PR #626, opened 2020-08-14
Graphql schema loading with headers

tsiq-karold:
We need headers from the header provider to be sent to the GraphQL introspection call.  Currently this does not happen because the http call to introspect specifies `Header.EMPTY`.  Doing something similar to `HttpConfigurations. fullUrlWithoutEnabledCheck()` will not work as it results in an infinite loop.  This is the best I've found so far.


tsiq-karold:
Closing while I work on a proper test and hopefully better fix.

PR #633, opened 2020-08-17
graphql: support authenticated servers

tsiq-karold:


webtau-groovy/src/main/groovy/org/testingisdocumenting/webtau/cfg/GroovyConfigBasedHttpConfiguration.groovy
```diff
@@ -49,7 +50,9 @@ class GroovyConfigBasedHttpConfiguration implements HttpConfiguration {
         }
 
         return HttpConfigurations.withDisabledConfigurations {
-            return headerProvider.get().call(fullUrl, passedUrl, given) as HttpHeader
+            HttpValidationHandlers.withDisabledHandlers {
```
MykolaGolubyev, line 53: Now I wonder if we are going to have the same issue in Java. Could you please add java full example in a separate PR when you have time?

webtau-http/src/main/java/org/testingisdocumenting/webtau/http/validation/HttpValidationHandlers.java
```diff
@@ -26,6 +26,7 @@
 public class HttpValidationHandlers {
     private static final List<HttpValidationHandler> globalHandlers = ServiceLoaderUtils.load(HttpValidationHandler.class);
     private static final ThreadLocal<List<HttpValidationHandler>> localHandlers = ThreadLocal.withInitial(ArrayList::new);
+    private static final ThreadLocal<Boolean> enabled = ThreadLocal.withInitial(() -> true);
```
MykolaGolubyev, line 29: missing copyright update

webtau-test-server/src/main/java/org/testingisdocumenting/webtau/http/testserver/GraphQLResponseHandler.java
```diff
@@ -76,11 +111,19 @@ public void handle(String url, Request baseRequest, HttpServletRequest request,
         } else {
             response.setStatus(405);
         }
+    }
 
-        baseRequest.setHandled(true);
+    private boolean isAuthorised(HttpServletRequest request) {
```
MykolaGolubyev, line 116: I know header is called "authorization", but I think what is happening here is Authentication?

tsiq-karold: Yes, it's weird and confusing but it's not my invention: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization

webtau-groovy/src/main/groovy/org/testingisdocumenting/webtau/cfg/GroovyConfigBasedHttpConfiguration.groovy
tsiq-karold: I created an equivalent Java test and we definitely have the same problem :(

tsiq-karold: I'll do that in a separate PR.

PR #634, opened 2020-08-18
core: introduce persona concept

MykolaGolubyev:


webtau-core-groovy/src/main/groovy/org/testingisdocumenting/webtau/persona/PersonaExtensions.groovy
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+package org.testingisdocumenting.webtau.persona
+
+import java.util.function.Supplier
+
+class PersonaExtensions {
+    static <R> R call(Persona persona, Closure code) {
+        return persona.execute(code as Supplier<R>)
```
tsiq-karold, line 23: Why a different name?  In the http extensions at least we've preserved the names.

webtau-core/src/main/java/org/testingisdocumenting/webtau/persona/Persona.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+
+package org.testingisdocumenting.webtau.persona;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.function.Supplier;
+
+public class Persona {
+    private static final ThreadLocal<Persona> currentPersona = new ThreadLocal<>();
```
tsiq-karold, line 24: If I understand this right, the thread local is purely to guard against nesting?  I suggest we add a comment explaining that for two reasons:
1. it's a bit different to most of our usage of thread locals which is for keeping and accessing global-ish state
1. it's not immediately obvious why we need it given that all usage of this class is against an explicitly created instance rather than through static functions.

webtau-core-groovy/src/main/groovy/org/testingisdocumenting/webtau/persona/PersonaExtensions.groovy
MykolaGolubyev: not following

webtau-core/src/main/java/org/testingisdocumenting/webtau/persona/Persona.java
MykolaGolubyev: ThreadLocal is to be able to have active contexts in a multithreaded execution. So test1 running in thread1 can have its context1 and the same for thread2. 
In my tests I plan to reuse persona instances but it is not required. It is similar to page objects in UI module in a way that you can either pre-create them or create on demand.

webtau-core-groovy/src/main/groovy/org/testingisdocumenting/webtau/persona/PersonaExtensions.groovy
tsiq-karold: The `call` vs `execute` thing.

webtau-core/src/main/java/org/testingisdocumenting/webtau/persona/Persona.java
tsiq-karold: I forgot about the ability to run tests in parallel.

webtau-core-groovy/src/main/groovy/org/testingisdocumenting/webtau/persona/PersonaExtensions.groovy
MykolaGolubyev: `call` is a groovy magic method -- this is what let's you do 
```
John {
}
```

tsiq-karold: Ah, ok

PR #636, opened 2020-08-18
build: leave clean working directory

tsiq-karold:
It's nice not to see a load of stuff in `git status` when you've not changed anything but have ran tests.  I've therefore added `doc-artifacts` to gitignore and also made one test generate its artifacts into that dir.

Ideally all these transient things that get generated by the build would go into `target` (and equivalent for npm) so that a clean removes them but that's a bit more work, this first step is easy and does a large part of the job.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ webtau-report-*.html
 .webtau.cache.json
 .vertx
 todo.md
+doc-artifacts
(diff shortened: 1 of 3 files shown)


MykolaGolubyev:
Some already do go to `target`. The problem I find with that approach (not saying that we shouldn't do it) is then I can't find the generated files using IDEA as it completely ignores `target` as it should.

tsiq-karold:
> Some already do go to `target`. The problem I find with that approach (not saying that we shouldn't do it) is then I can't find the generated files using IDEA as it completely ignores `target` as it should. 

That is true. Alternative is to add something custom to maven's clean goal.

PR #639, opened 2020-08-19
browser: browser instance per persona per thread

MykolaGolubyev:


webtau-feature-testing/examples/scenarios/ui/webtau.persona.cfg.groovy
```diff
@@ -0,0 +1,8 @@
+package scenarios.ui
+
+url = "http://localhost:8180"
+
+personas {
+    John {
```
MykolaGolubyev, line 6: this is a placeholder for now, need to figure out how to deal with Persona specific configs. Things come to mind - browser size/placement. DB username/password. HTTP headers(?)
@tsiq-karold

PR #648, opened 2020-08-26
graphql: full set of execute overloads

### LINKED ISSUE #603: GraphQL DSL - full set of execute overloads
Should support all valid combinations of:
* query
* variables
* operation name
* headers
* validators (or lack of)

tsiq-karold:
This wasn't at all tedious 😁 

Fixes #603

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ public class WeatherIT {
     @Test
     public void checkWeather() {
         graphql.execute("{ weather { temperature } }", (header, body) -> {
-            body.get("weather.temperature").shouldBe(lessThan(100));
+            body.get("data.weather.temperature").shouldBe(lessThan(100));
         });
     }
 }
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -91,7 +91,7 @@ public class WeatherIT {
     @Test
     public void checkWeather() {
         graphql.execute("{ weather { temperature } }", (header, body) -> {
-            body.get("weather.temperature").shouldBe(lessThan(100));
+            body.get("data.weather.temperature").shouldBe(lessThan(100));
```
MykolaGolubyev, line 94: is `data.` optional here?

tsiq-karold: No, it exists in the response so when interacting with the body DataNode directly you need it.  The optionality of it is implemented in the resolution customisation when we convert the groovy closure into the relevant validator interface.

Issue #638, opened 2020-08-18
Swagger - error when asserting on fields not in schema


tsiq-karold:
If one is asserting that a field is null and makes a typo in the field name then the test will always pass even if the field comes back non-null.

We should validate that the assertion is being made against a field which is actually part of the schema. We will also need to provide a way to turn off this behaviour should someone really want to assert something that does not exist in the schema. e.g.
```groovy
http.get('/foo') {
  suppressUnknownFieldErrors {
    somethingNotInSchema.should == 'foo'
  }
}
```

Related to #637 but we probably cannot handle this generically in HTTP because there may not be a schema.

MykolaGolubyev:
do you still need this one?

tsiq-karold:
I don't need it as such but I do think it would be a useful addition to webtau.

MykolaGolubyev:
I recall you had some implementation around it, is it still somewhere there?

tsiq-karold:
I do not recall that :)

MykolaGolubyev:
closing as not active

Issue #637, opened 2020-08-18
GraphQL - error when asserting on fields not in schema


tsiq-karold:
If one is asserting that a field is null and makes a typo in the field name then the test will always pass even if the field comes back non-null.

We should validate that the assertion is being made against a field which is actually part of the schema.  In GraphQL, this will require us to fetch more data via introspection because we need to check for field names in the return objects rather than in the query.

We will also need to provide a way to turn off this behaviour should someone really want to assert something that does not exist in the schema.  e.g.
```groovy
graphql.execute(query) {
  suppressUnknownFieldErrors {
    somethingNotInSchema.should == 'foo'
  }
}
```

MykolaGolubyev:
Nice

tsiq-karold:
> @tsiq-karold any remaining motivation for this?

Not right now, no.

MykolaGolubyev:
closing as not active

PR #658, opened 2020-09-02
core: data node id validators

tsiq-karold:
Intention behind these validators is to use them to ensure that assertions are made on fields that exist in schema, whether that be OpenAPI (#638) or GraphQL (#637).

I'm not completely crazy about the `DataNodeIdValidator` name but struggling for better alternatives, happy to hear your thoughts on this @MykolaGolubyev .  Part of the reason I don't like it is when I add the ability to bypass it, it would look like this and the naming isn't great:
```groovy
http.get("/foo") {
  DataNodeIdValidators.withoutFieldSchemaValidation {
    body.doesnotexist.should == null
  }
}
```


MykolaGolubyev:
closing PR for now

PR #665, opened 2020-09-04
core: track null data nodes

tsiq-karold:

PR #667, opened 2020-09-06
core: split structured data node and track null nodes

tsiq-karold:
@MykolaGolubyev as discussed, I tried it :) and I think it's better.  Let me know what you think.

### DIFF
--- a/webtau-http/src/main/java/org/testingisdocumenting/webtau/http/datanode/DataNode.java
+++ b/webtau-http/src/main/java/org/testingisdocumenting/webtau/http/datanode/DataNode.java
@@ -1,4 +1,5 @@
 /*
+ * Copyright 2020 webtau maintainers
  * Copyright 2019 TWO SIGMA OPEN SOURCE, LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -21,35 +22,59 @@ import org.testingisdocumenting.webtau.expectation.ActualPath;
 import org.testingisdocumenting.webtau.expectation.equality.CompareToComparator;
 import org.testingisdocumenting.webtau.expectation.equality.CompareToResult;
 
+import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
 import static org.testingisdocumenting.webtau.WebTauCore.createActualPath;
 
-public interface DataNode extends DataNodeExpectations, Comparable, Iterable<DataNode> {
+public interface DataNode extends DataNodeExpectations, Comparable<Object>, Iterable<DataNode> {
     DataNodeId id();
 
     DataNode get(String pathOrName);
 
-    boolean has(String pathOrName);
+    default boolean has(String pathOrName) {
+        return !get(pathOrName).isNull();
+    }
 
     DataNode get(int idx);
 
-    TraceableValue getTraceableValue();
+    default TraceableValue getTraceableValue() {
+        return null;
+    }
 
     <E> E get();
 
-    boolean isList();
+    default boolean isList() {
+        return false;
+    }
 
-    boolean isSingleValue();
+    default boolean isSingleValue() {
+        return false;
+    }
 
-    List<DataNode> elements();
+    default List<DataNode> elements() {
+        return Collections.emptyList();
+    }
 
-    int numberOfChildren();
+    @Override
+    @SuppressWarnings("NullableProblems")
+    default Iterator<DataNode> iterator() {
+        return elements().iterator();
+    }
 
-    int numberOfElements();
+    default int numberOfChildren() {
+        return 0;
+    }
 
-    Map<String, DataNode> asMap();
+    default int numberOfElements() {
+        return 0;
+    }
+
+    default Map<String, DataNode> asMap() {
+        return Collections.emptyMap();
+    }
 
     default boolean isNull() {
         return false;
@@ -67,6 +92,7 @@ public interface DataNode extends DataNodeExpectations, Comparable, Iterable<Dat
     }
 
     @Override
+    @SuppressWarnings("NullableProblems")
     default int compareTo(Object rhv) {
         CompareToComparator comparator = CompareToComparator.comparator(
                 CompareToComparator.AssertionMode.GREATER_THAN);
--- a/webtau-http/src/main/java/org/testingisdocumenting/webtau/http/datanode/DataNodeCompareToHandler.java
+++ b/webtau-http/src/main/java/org/testingisdocumenting/webtau/http/datanode/DataNodeCompareToHandler.java
@@ -16,12 +16,19 @@
 
 package org.testingisdocumenting.webtau.http.datanode;
 
+import org.testingisdocumenting.webtau.data.traceable.CheckLevel;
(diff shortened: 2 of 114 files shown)


webtau-http/src/main/java/org/testingisdocumenting/webtau/http/datanode/DataNode.java
```diff
@@ -21,35 +22,59 @@
...
+        return 0;
+    }
 
-    int numberOfElements();
+    default int numberOfElements() {
+        return 0;
+    }
 
-    Map<String, DataNode> asMap();
+    default Map<String, DataNode> asMap() {
```
MykolaGolubyev, line 75: have mixed feelings about these defaults. It feels like it is not really a default implementation, but more like you know 2 out of 3 implementations are similar and hence they are there. Maybe it is OK in these case since we are not going to have more implementations.

webtau-http/src/main/java/org/testingisdocumenting/webtau/http/datanode/ListDataNode.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @Override
+    public DataNodeId id() {
+        return id;
+    }
+
+    @Override
+    public DataNode get(String nameOrPath) {
+        if (values.stream().noneMatch(v -> v.has(nameOrPath))) {
+            return new NullDataNode(id.child(nameOrPath));
```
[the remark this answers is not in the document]
tsiq-karold: I think I agree but this is not new, this is how it currently is, just moved the code.

webtau-http/src/main/java/org/testingisdocumenting/webtau/http/datanode/DataNode.java
tsiq-karold: Yes, I was equally in two minds.  I did it not so much for removing duplication but rather for decluttering the specific impls.  For example, the list impl only worries about functions relevant to list, not all this other stuff.

MykolaGolubyev: In a unlikely event that someone will implement DataNode interface I would only left real defaults here.

webtau-http/src/main/java/org/testingisdocumenting/webtau/http/datanode/ListDataNode.java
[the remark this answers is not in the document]
MykolaGolubyev: So this is good that DataNode is getting a face lift. I think it never had a second iteration! Thanks for doing it.

webtau-http/src/main/java/org/testingisdocumenting/webtau/http/datanode/DataNode.java
tsiq-karold: There was more I initially had as defaults but held back a bit.  Which ones did you have in mind?

tsiq-karold:
@MykolaGolubyev as discussed I've pushed the tracking of null nodes to this branch too.  It is as you predicted a bit duplicative.  I still think the split has some value just from the logical reasoning perspective if nothing else.

We have a few options:
1. don't split and I'll finish up the null tracking in #665 
1. leave as is now in this PR
1. create an abstract type to sit between `DataNode` and the List/Map/Value DataNodes which has the `queriedNonExistentNodes` map plus a function to save repeating the `computeIfAbsent` in many places

MykolaGolubyev:
3. I would replace abstract with some sort of composite trait that can be used in each of the classes. 
On top of adding similar code to each class, I think the tests will also have to be repeated.

webtau-http/src/test/groovy/org/testingisdocumenting/webtau/http/datanode/ValueDataNodeTest.groovy
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        checkLevelIsRecordedForNonExistentNode {
+            node["key"]
+        }
+
+        checkLevelIsRecordedForNonExistentNode {
+            node.get(0)
+        }
+    }
```
MykolaGolubyev, line 79: I tried this
```
def node = DataNodeBuilder.fromValue(new DataNodeId("body"), null)

code {
    node.should == [1, [key: 'value']]
} should throwException(~/mismatches/)

node.get(0).get('key').traceableValue.checkLevel.should == CheckLevel.ExplicitFailed
```
it failed. Is this expected behavior?

tsiq-karold:
>I would replace abstract with some sort of composite trait that can be used in each of the classes.

I was about to do that but I think it'll just replace the map with an instance of something else and calls to `computeIfAbsent` with a marginally simpler method call.  Not sure that in itself is worth it.  Having said, as we look at the sort of stuff I started with #658 and also data coverage, we may have more functionality to put into this class so may actually be worth doing now.

webtau-http/src/test/groovy/org/testingisdocumenting/webtau/http/datanode/ValueDataNodeTest.groovy
tsiq-karold: I don't think so.  I'll try it.

tsiq-karold: Having just read your test closer, I think this is expected.  You're asserting that `node` is `[1, [key: 'value']]` which should mark `node.get(0)` and `node.get(1).key` as checked yet your check level test is looking at `node.get(0).key`.

Having said that, the following assertion also fails:
```groovy
node.get(1).get('key').traceableValue.checkLevel.should == CheckLevel.ExplicitFailed
```

tsiq-karold: This commit should address this: https://github.com/testingisdocumenting/webtau/pull/667/commits/a3d9716b482d87711bbd89c02fe85cbb4f2984e1

webtau-http/src/main/java/org/testingisdocumenting/webtau/http/datanode/DataNodeCompareToHandler.java
```diff
@@ -41,21 +48,33 @@ public boolean handleGreaterLessEqual(Object actual, Object expected) {
...
-            compareWithMap(comparator, actualPath, (DataNode) actual, (Map) expected);
+            result = compareWithMap(comparator, actualPath, actualDataNode, (Map) expected);
         } else {
-            Object extractedActual = extractActual((DataNode) actual);
-            comparator.compareUsingEqualOnly(actualPath, extractedActual, expected);
+            Object extractedActual = extractActual(actualDataNode);
+            result = comparator.compareUsingEqualOnly(actualPath, extractedActual, expected);
+        }
+
+        if (actualDataNode.isNull() || actualDataNode.get() == null) {
```
MykolaGolubyev, line 60: what if actual is not directly null, but something like 
```
person.should == [address: [street: 'name']
```
where `address` is not present on actual `person`

MykolaGolubyev:
closing for now

PR #675, opened 2020-09-15
http: set proxy via config

MykolaGolubyev:


webtau-config/src/main/java/org/testingisdocumenting/webtau/cfg/WebTauConfig.java
```diff
@@ -48,6 +48,11 @@
     private final ConfigValue config = declare("config", "config file path", () -> CONFIG_FILE_NAME_DEFAULT);
     private final ConfigValue env = declare("env", "environment id", () -> "local");
     private final ConfigValue url = declare("url", "base url for application under test", NO_DEFAULT);
+
+    private final ConfigValue httpProxy = declare("httpProxy", "http proxy host:port", NO_DEFAULT);
+    private final ConfigValue httpsProxy = declare("httpsProxy", "https proxy host:port", NO_DEFAULT);
+    private final ConfigValue noProxy = declare("noProxy", "list of hosts to exclude from proxy host1|*.host2|host3.*", NO_DEFAULT);
```
tsiq-karold, line 54: How about `noProxyHosts`?  If I saw `noProxy` without the description, I'd assume it was a boolean toggle, it's not clear this is a list of hosts from the name.

webtau-config/src/main/java/org/testingisdocumenting/webtau/cfg/WebTauCoreConfigHandler.java
```diff
@@ -37,5 +37,47 @@
...
+        setNoProxy(cfg);
+    }
+
+    private void setProxy(String httpPrefix, ConfigValue configValue) {
+        if (configValue.isDefault()) {
+            return;
+        }
+
+        HostPort hostPort = new HostPort(configValue.getAsString());
+        System.setProperty(httpPrefix + "." + "proxyHost", hostPort.host);
```
tsiq-karold, line 56: Can you remind me exactly when the config handling would happen in a Junit test?  IIRC the proxy setup is essentially a lazily initialised singleton so if there was any way for someone to invoke some code, prior to cfg handling, which ends up initialising http proxies, this property setting here won't have any effect.

webtau-feature-testing/src/test/groovy/org/testingisdocumenting/webtau/featuretesting/WebTauRestFeaturesTest.groovy
```diff
@@ -147,6 +147,11 @@ class WebTauRestFeaturesTest {
         runCli('springboot/listMatchByKey.groovy', 'springboot/webtau.cfg.groovy', "--url=$customersBaseUrl")
     }
 
+    @Test
+    void "proxy config"() {
+        runCli("proxy/passingProxy.groovy", "proxy/webtau.proxy.cfg.groovy", "--url=${testRunner.testServer.uri}")
```
tsiq-karold, line 152: How does this actually work?  Any tests after this one will have those system properties set so there is a possibility they might actually run with a proxy.  I think this goes back to my earlier point around the lazy initialisation of the proxy and maybe we're actually relying on that here.

webtau-config/src/main/java/org/testingisdocumenting/webtau/cfg/WebTauConfig.java
MykolaGolubyev: thanks, makes sense

MykolaGolubyev:
related to #948

PR #678, opened 2020-09-20
http: expose datanode.children() method

MykolaGolubyev:


tsiq-karold (commented):
What's the use case for accessing the list of children directly?  Wouldn't `node.toMap().values()` do?

MykolaGolubyev:
Good point. toMap feels like impl details for some reason. 
Basically use case is to do iteration and validation for all children.

tsiq-karold:
We could in theory repurpose the `elements()` method for this but I'm not sure it's a good idea.  If we do add a `getChildren()` I think we should either wrap in an immutable map or return just a stream.

As a related aside, I actually don't like the existing `toMap()` impl because the side effects of `node.toMap().get("foo")` or `node.toMap().has("foo")` for that matter are different to `node.get("foo")` or `node.has("foo")`.  Right now this is harmless but has negative consequences for my null tracking PR and my next commit there will likely change the `toMap()` behaviour a bit.

MykolaGolubyev:
What do you think of hiding toMap and change internal code to use children stream or immutable list?

tsiq-karold:
Makes sense.  I doubt anyone is relying on `toMap()` though if they are we will obviously break them.

MykolaGolubyev:
@tsiq-karold please take another look

webtau-http-groovy/src/main/groovy/org/testingisdocumenting/webtau/http/datanode/GroovyDataNode.groovy
```diff
@@ -91,6 +92,11 @@ class GroovyDataNode implements DataNodeExpectations, DataNode {
         return node.elements().collect { new GroovyDataNode(it) }
     }
 
+    @Override
+    Collection<DataNode> children() {
```
tsiq-karold, line 96: Should this be a stream so it cannot be modified by callers?

webtau-http/src/main/java/org/testingisdocumenting/webtau/http/datanode/DataNodeCompareToHandler.java
```diff
@@ -55,10 +56,10 @@ public void compareGreaterLessEqual(CompareToComparator compareToComparator, Act
         compareToComparator.compareUsingCompareTo(actualPath, actualDataNode.getTraceableValue(), expected);
     }
 
-    private void compareWithMap(CompareToComparator comparator, ActualPath actualPath, DataNode actual, Map expected) {
-        Map<String, DataNode> actualAsMap = actual.asMap();
+    private void compareWithMap(CompareToComparator comparator, ActualPath actualPath, DataNode actual, Map<?, ?> expected) {
+        Map<String, DataNode> actualAsMap = DataNodeToMapConverter.convertToMap(actual);
```
tsiq-karold, line 60: I think instead of converting to a map here, we should do this:
```java
    private void compareWithMap(CompareToComparator comparator, ActualPath actualPath, DataNode actual, Map expected) {
        Set keys = expected.keySet();
        for (Object key : keys) {
            String p = (String) key;
            ActualPath propertyPath = actualPath.property(p);

            Object expectedValue = expected.get(p);
            if (! actual.has(p)) {
                comparator.reportMissing(this, propertyPath, expectedValue);
            } else {
                comparator.compareUsingEqualOnly(propertyPath, actual.get(p), expectedValue);
            }
        }
    }
```
Then all operations are directly on the node therefore any future null tracking (😁 ) will be aware of these access attempts.

webtau-http/src/main/java/org/testingisdocumenting/webtau/http/datanode/DataNodeCompareToHandler.java
```diff
@@ -89,6 +90,6 @@ private Object extractActual(DataNode actual) {
             return actual.elements();
         }
 
-        return actual.asMap();
+        return DataNodeToMapConverter.convertToMap(actual);
```
tsiq-karold, line 93: Ideally this would also convert to something that would track `has` and `get` calls on the node itself.  Does the caller of this function really need a map or could they work with the data node directly?

webtau-http/src/main/java/org/testingisdocumenting/webtau/http/datanode/DataNodeCompareToHandler.java
MykolaGolubyev: good idea

webtau-http-groovy/src/main/groovy/org/testingisdocumenting/webtau/http/datanode/GroovyDataNode.groovy
MykolaGolubyev: It is protected for modification, not as implicit as `elements()`, but you can't change it. 
I would prefer to leave it as `Collection` to let folks, including us, do a good old `for(el in list) {}` loop

webtau-http/src/main/java/org/testingisdocumenting/webtau/http/datanode/DataNodeCompareToHandler.java
MykolaGolubyev: updated code and added comment and a test

webtau-http-groovy/src/main/groovy/org/testingisdocumenting/webtau/http/datanode/GroovyDataNode.groovy
tsiq-karold: What protects it from modification?  `StructuredDataNode.children()` returns `children.values()` and looking at the impl of `LinkedHashMap` for example I don't think it is protected from modification.  It seems to not allow insertions/updates but odly enough allows removals.  I tried this to be sure I'm not completely misunderstanding the code, this was my rather unsophisticated test:
```java
    public static void main(String[] args) {
        LinkedHashMap<String, String> map = new LinkedHashMap<>();
        map.put("foo", "bar");
        System.out.println(map);

        map.values().clear();
        System.out.println(map);
    }
```
Note the `values().clear()`, not direct `clear()` on the map.  This printed:
```
{foo=bar}
{}
```

MykolaGolubyev: Oh, wow, thanks for checking - was under impression that it doesn't allow. Will wrap into Collections.unmodifiable

tsiq-karold: I also assumed that it was essentially a view.

tsiq-karold (approved):

PR #688, opened 2020-10-12
persona: disallow empty persona name

tsiq-karold:


webtau-core/src/main/java/org/testingisdocumenting/webtau/expectation/code/ThrowExceptionMatcher.java
```diff
@@ -103,6 +103,10 @@ private void extractExceptionDetails(Throwable t) {
             Throwable undeclaredCheckedException = t.getCause();
             thrownMessage = undeclaredCheckedException.getMessage();
             thrownClass = undeclaredCheckedException.getClass();
+        } else if (t instanceof ExceptionInInitializerError) {
```
MykolaGolubyev, line 106: oO

webtau-core/src/main/java/org/testingisdocumenting/webtau/persona/Persona.java
```diff
@@ -16,24 +16,35 @@
...
 
     public Persona(String id) {
         this(id, Collections.emptyMap());
     }
 
     public Persona(String id, Map<String, Object> payload) {
+        this(id, payload, false);
+    }
+
+    private Persona(String id, Map<String, Object> payload, boolean isDefault) {
```
MykolaGolubyev, line 41: oh, this may conflict with the things in the other PR I have. I have `DEFAULT_PERSONA_ID` defined there...

webtau-core/src/main/java/org/testingisdocumenting/webtau/expectation/code/ThrowExceptionMatcher.java
tsiq-karold: ?

webtau-core/src/main/java/org/testingisdocumenting/webtau/persona/Persona.java
tsiq-karold: Yeah, I didn't want to add this isDefault property really.  However, I couldn't just have the id is null or empty check in the c'tor because that resulted in the default persona creation throwing 😁

webtau-core/src/main/java/org/testingisdocumenting/webtau/expectation/code/ThrowExceptionMatcher.java
MykolaGolubyev: surprised to see this, that's all. Fair change. And also signals that we may need to throw at a different time maybe to make exception more user friendly? Like maybe another check in the `persona` method?

webtau-core/src/main/java/org/testingisdocumenting/webtau/persona/Persona.java
MykolaGolubyev: sorry I didn't notice it is a private constructor, should be good.

webtau-core/src/test/java/org/testingisdocumenting/webtau/persona/PersonaTest.java
```diff
@@ -61,4 +61,14 @@ public void shouldNotAllowNestingPersonas() {
         }).should(throwException("nesting personas is not allowed, active persona id: John, " +
                 "attempted to nest persona id: Bob"));
     }
-}
\ No newline at end of file
+
+    @Test
+    public void cannotCreateAPersonWithSameNameAsDefaultPersona() {
```
MykolaGolubyev, line 66: for consistency, do you think we should add test to `PersonaExtensionsTest` as well? The reason I have some tests in there is to make sure we don't assume that groovy DSL delegates strictly to Java. But not sure it makes sense in this case.

webtau-core/src/main/java/org/testingisdocumenting/webtau/expectation/code/ThrowExceptionMatcher.java
tsiq-karold: Yeah, I considered making the c'tor private and adding factory methods instead which would have avoided this.  I don't feel too strongly either way.

MykolaGolubyev: I meant the factory method we already have and which most users will probably use. But then we need double guarding. I like your suggestion better

PR #693, opened 2020-10-15
persona: add persona column to http calls tab

### LINKED ISSUE #691: HTTP report: Include persona name in HTTP calls tab
Currently it's only visible in the Steps tab.

MykolaGolubyev:
closes #691 
image

image

### DIFF
--- a/webtau-http/src/main/java/org/testingisdocumenting/webtau/http/validation/HttpValidationResult.java
+++ b/webtau-http/src/main/java/org/testingisdocumenting/webtau/http/validation/HttpValidationResult.java
@@ -202,7 +202,11 @@ public class HttpValidationResult implements TestStepPayload {
         Map<String, Object> result = new LinkedHashMap<>();
 
         result.put("id", id);
-        result.put("personaId", personaId);
+
+        if (!personaId.isEmpty()) {
+            result.put("personaId", personaId);
+        }
+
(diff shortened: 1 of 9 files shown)


webtau-http/src/main/java/org/testingisdocumenting/webtau/http/validation/HttpValidationResult.java
```diff
@@ -197,6 +202,11 @@ public long getElapsedTime() {
         Map<String, Object> result = new LinkedHashMap<>();
 
         result.put("id", id);
+
+        if (!personaId.isEmpty()) {
```
tsiq-karold, line 206: Should we pass the whole `Persona` to `HttpValidationResult` so that this check here can be `!persona.isDefault()` so we don't propagate knowledge of what ID constitutes the default persona?  At a minimum I think this should compare to the `DEFAULT_PERSONA_ID` constant rather than checking it's empty.

MykolaGolubyev: You are right. On report side I still rely on it to be empty though. Adding a change. 
I read it rather as "persona is not set" vs "persona is default". Well I read it both ways depending on the context 🤷‍♂️

tsiq-karold (approved):

PR #698, opened 2020-10-19
maven plugin: sort test files

tsiq-karold:
The order doesn't seem totally deterministic, at least not across OSes, so I suggest we sort alphabetically.


MykolaGolubyev (approved):
makes sense. Thank you

PR #699, opened 2020-10-19
graphql: update docs to clarify queries and mutations

### LINKED ISSUE #697: GraphQL queries/mutation docs clarification
We have page called `Queries` but covers queries and mutations. We either rename it to to avoid accidental skipping through it. Or have a separate pages to cover mutations

tsiq-karold:
Fixes #697

### DIFF
--- a/webtau-docs/znai/GraphQL/queries-and-mutations.md
+++ b/webtau-docs/znai/GraphQL/queries-and-mutations.md
@@ -1,11 +1,11 @@
-# Executing Queries
+# Executing Queries and Mutations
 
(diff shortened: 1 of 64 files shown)


MykolaGolubyev (approved):

webtau-docs/znai/GraphQL/queries-and-mutations.md
```diff
@@ -1,11 +1,11 @@
-# Executing Queries
+# Executing Queries and Mutations
```
MykolaGolubyev, line 1: Actually should we have two sections here. One for queries and one for mutations?

tsiq-karold: I don't think so, there's no difference in how you execute them or how you'd assert

PR #701, opened 2020-10-23
repl: multiple scenarios selection range and list

MykolaGolubyev:


webtau-groovy/src/main/groovy/org/testingisdocumenting/webtau/cli/repl/ReplScenariosSelector.groovy
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        if (selector instanceof Map) {
+            return selectSingleScenariosByRange(selector)
+        }
+
+        return [new SelectInputAndResult(selector, -1, null)]
+    }
+    
+    private List<SelectInputAndResult> selectSingleScenarioByRegexp(String regexp) {
```
tsiq-karold, line 53: Why do all the `selectSingle...` methods return a list? That seems contradictory to the whole single thing :)

webtau-groovy/src/main/groovy/org/testingisdocumenting/webtau/cli/repl/ReplScenariosSelector.groovy
```diff
@@ -0,0 +1,125 @@
...
+    private List<SelectInputAndResult> selectSingleScenariosByRange(Map range) {
+        if (range.size() != 1) {
+            return [new SelectInputAndResult(range, -1, null)]
+        }
+
+        def entrySet = range.iterator().next()
+        return selectSingleScenariosByRange(entrySet.key, entrySet.value)
+    }
+
+    private List<SelectInputAndResult> selectSingleScenariosByRange(Object from, Object to) {
```
tsiq-karold, line 77: This one actually returns more than one thing so it's not "single".

MykolaGolubyev: Thanks for catching. Yes this is not single. The other ones return lists so it is easier to combine in one go and flatten

tsiq-karold: OK, it's a bit weird to return a list which always has one thing.  Looking at the usage we just call these things from a big if/else if/else where it would be easy to just warp the single element in a list.  I would personally find that less confusing (also if we have tests for this class directly, you won't have to test the list is a single item list :) ).  That said, it's not a huge deal so I won't block the PR on it.

tsiq-karold (approved):

PR #716, opened 2020-11-04
remove graphQLEnabled flag

### LINKED ISSUE #702: GraphQL - remove the enabled config flag
We should be able to detect by trying an introspection call and then set it enabled or disabled accordingly.

tsiq-karold:
I removed the `graphQLEnabled` flag but I have replaced it with `graphQLIgnoreIntrospectionFailures` so that people who want to be aware of such failures can be.

Fixes #702


MykolaGolubyev (approved):
Thank you

PR #741, opened 2020-12-10
build: build github action

MykolaGolubyev:


.github/workflows/ci-build.yaml
```diff
@@ -0,0 +1,26 @@
+name: Build znai
```
tsiq-karold, line 1: Hmm, this seems like the wrong project name 😁

.github/workflows/ci-build.yaml
```diff
@@ -0,0 +1,26 @@
...
+        with:
+          java-version: ${{ matrix.java }}
+
+      - name: Set up grahpviz
+        run: sudo apt-get install graphviz
+
+      - name: Build with xvfb enabled
+        uses: GabrielBB/xvfb-action@v1.0
+        with:
+          run: mvn -B package
```
tsiq-karold, line 26: I would suggest `verify` rather than `package` in case we ever add plugins that run during `verify`.

PR #768, opened 2021-01-09
graphql: graphql listener

### LINKED ISSUE #766: GraphQL - implement a GraphQLListener
Similar to `HttpListener`, implement a GraphQL listener so GraphQL users do not need to parse http requests and responses in an `HttpListener`.

tsiq-karold:
Fixes #766

### DIFF
--- a/webtau-graphql-groovy/src/test/groovy/org/testingisdocumenting/webtau/graphql/GraphQLListenersTest.groovy
+++ b/webtau-graphql-groovy/src/test/groovy/org/testingisdocumenting/webtau/graphql/GraphQLListenersTest.groovy
@@ -69,7 +69,7 @@ class GraphQLListenersTest extends GraphQLTestBase implements GraphQLListener {
     }
 
     @Test
-    void "before and after query callbacks should have correct response data"() {
+    void "after query callback should have correct response data"() {
         graphql.execute(ERROR_QUERY, [msg: "test error msg"])
 
         afterPayload.data.should == [error: null]
@@ -78,7 +78,7 @@ class GraphQLListenersTest extends GraphQLTestBase implements GraphQLListener {
     }
 
     @Test
-    void "remove removes both graphql and http listeners"() {
+    void "removed listeners are not invoked"() {
         cleanup()
 
         graphql.execute(QUERY)
--- a/webtau-graphql-groovy/src/test/groovy/org/testingisdocumenting/webtau/graphql/GraphQLListenersTest.groovy
+++ b/webtau-graphql-groovy/src/test/groovy/org/testingisdocumenting/webtau/graphql/GraphQLListenersTest.groovy
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2021 webtau maintainers
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.testingisdocumenting.webtau.graphql
 
 import org.junit.After
--- a/webtau-graphql-groovy/src/test/groovy/org/testingisdocumenting/webtau/graphql/GraphQLListenersTest.groovy
+++ b/webtau-graphql-groovy/src/test/groovy/org/testingisdocumenting/webtau/graphql/GraphQLListenersTest.groovy
@@ -35,7 +35,7 @@ class GraphQLListenersTest extends GraphQLTestBase implements GraphQLListener {
             variables    : variables,
             operationName: operationName,
             requestHeader: requestHeader,
-            data         : data,
+            data         : responseData,
             errors       : errors
         ]
     }
--- a/webtau-graphql-groovy/src/test/groovy/org/testingisdocumenting/webtau/graphql/GraphQLListenersTest.groovy
+++ b/webtau-graphql-groovy/src/test/groovy/org/testingisdocumenting/webtau/graphql/GraphQLListenersTest.groovy
@@ -1,0 +1,100 @@
+package org.testingisdocumenting.webtau.graphql
+
+import org.junit.After
+import org.junit.Before
+import org.junit.Ignore
+import org.junit.Test
+import org.testingisdocumenting.webtau.graphql.listener.GraphQLListener
+import org.testingisdocumenting.webtau.graphql.listener.GraphQLListeners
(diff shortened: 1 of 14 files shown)


webtau-graphql/src/main/java/org/testingisdocumenting/webtau/graphql/listener/GraphQLListeners.java
```diff
@@ -0,0 +1,87 @@
...
+        listeners.stream()
+                .filter(l -> l.listener == listener)
+                .findFirst()
+                .ifPresent(l -> {
+                    HttpListeners.remove(l);
+                    listeners.remove(l);
+                });
+    }
+
+    private static class WrappedGraphQLListener implements GraphQLListener, HttpListener {
```
tsiq-karold, line 50: I chose to do this wrapping approaching because it was the most elegant way to actually get access to the response.  I could have called the after callback from within `GraphQL.execute` but I only have a `DataNode` in there.

webtau-graphql-groovy/src/test/groovy/org/testingisdocumenting/webtau/graphql/GraphQLListenersTest.groovy
```diff
@@ -0,0 +1,96 @@
...
+    void "before and after query callbacks should have correct response data"() {
+        graphql.execute(ERROR_QUERY, [msg: "test error msg"])
+
+        afterPayload.data.should == [error: null]
+        afterPayload.errors[0].message.should == "Exception while fetching data (/error) : Error executing query: test error msg"
+        afterPayload.errors.size().should == 1
+    }
+
+    @Test
+    void "remove removes both graphql and http listeners"() {
```
MykolaGolubyev, line 81: remove removes?

webtau-graphql/src/main/java/org/testingisdocumenting/webtau/graphql/model/GraphQLRequest.java
```diff
@@ -18,15 +18,90 @@
...
+        Map<String, Object> variables = null;
+        Object variablesObj = request.get("variables");
+        if (variablesObj instanceof Map) {
+            variables = (Map<String, Object>) variablesObj;
+        } else if (variablesObj != null) {
+            // Ignoring as it's not a graphql request
+            return Optional.empty();
+        }
+
+        String operationName = null;
```
MykolaGolubyev, line 80: should we move these block as creation functions?

webtau-graphql/src/main/java/org/testingisdocumenting/webtau/graphql/listener/GraphQLListener.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    /**
+     * called after each graphql query
+     */
+    default void afterGraphQLQuery(String query,
+                                   Map<String, Object> variables,
+                                   String operationName,
+                                   HttpHeader requestHeader,
+                                   Map<String, Object> data,
```
MykolaGolubyev, line 31: how about renaming data to responseData?

webtau-graphql/src/main/java/org/testingisdocumenting/webtau/graphql/listener/GraphQLListeners.java
```diff
@@ -0,0 +1,87 @@
...
+        listeners.forEach(l -> l.listener.beforeGraphQLQuery(query, variables, operationName, requestHeader));
+    }
+
+    public static void add(GraphQLListener listener) {
+        listeners.add(new WrappedGraphQLListener(listener));
+    }
+
+    public static void remove(GraphQLListener listener) {
+        listeners.stream()
+                .filter(l -> l.listener == listener)
```
MykolaGolubyev, line 42: is this something `removeIf` can do? I may be missing something

webtau-graphql-groovy/src/test/groovy/org/testingisdocumenting/webtau/graphql/GraphQLListenersTest.groovy
```diff
@@ -0,0 +1,96 @@
...
+        afterPayload.data.should == [
+            'taskById': [
+                id: 'a'
+            ]
+        ]
+        afterPayload.errors.should == null
+    }
+
+    @Test
+    void "before and after query callbacks should have correct response data"() {
```
MykolaGolubyev, line 72: before and after and we check only after. the one above has similar naming

MykolaGolubyev (changes requested):
this is quite refactoring

webtau-graphql-groovy/src/test/groovy/org/testingisdocumenting/webtau/graphql/GraphQLListenersTest.groovy
tsiq-karold: Lol, what I meant is a call to remove listener actually removes it.  I'll reword.

webtau-graphql/src/main/java/org/testingisdocumenting/webtau/graphql/model/GraphQLRequest.java
tsiq-karold: There was a reason why I didn't do it but I can't remember what it is.  I'll have another look though.

webtau-graphql/src/main/java/org/testingisdocumenting/webtau/graphql/listener/GraphQLListener.java
tsiq-karold: Makes sense.

webtau-graphql/src/main/java/org/testingisdocumenting/webtau/graphql/listener/GraphQLListeners.java
tsiq-karold: No, because I have to also remove from HttpListeners.

webtau-graphql-groovy/src/test/groovy/org/testingisdocumenting/webtau/graphql/GraphQLListenersTest.groovy
tsiq-karold: The one above does check both but you're right this one is only one, I'll rename.

webtau-graphql/src/main/java/org/testingisdocumenting/webtau/graphql/model/GraphQLRequest.java
tsiq-karold: OK, I remember now.  There are two reasons I didn't stick it into functions:
1. behaviour for query is different to vars/operationName as the query is mandatory and the others are not.
2. there are three cases for vars/operationName.  It's null, it's not the expected type or it is the expected type.  Null and expected type result in me proceeding, not expected type results in not creating a GraphQLRequest as it's not a valid GraphQL request.  I think in all likelihood the usage of a function to deal with this will look almost identical to the current code and as the actual checks for type and null are very simple it didn't seem worthwhile deduplicating those parts alone.

MykolaGolubyev (approved):

PR #769, opened 2021-01-10
cli: explicit cfg path scan through; remove shell wrapper

MykolaGolubyev:


tsiq-karold (approved):
Code looks fine to me.  I wonder if this approach will get stuck at some point with commands similar to `nohup` which take the actual command to run as an arg but we can cross that bridge if we ever come to it.

PR #777, opened 2021-01-16
documentation: add brew install

MykolaGolubyev:
```
brew install testingisdocumenting/brew/webtau
```


MykolaGolubyev:
@tsiq-clemens @tsiq-karold @motlin FYI

tsiq-karold:
Nice, this is exciting :)

Does brew get automatically updated when we create releases?

MykolaGolubyev:
No :( And for now it is a brew formula that is in our own repo, not core repo. Either way it requires PR to update a version. Need to change url with zip and checksum.

PR #808, opened 2021-02-10
openapi: lazy init; relative http spec url

MykolaGolubyev:


MykolaGolubyev:
@tsiq-karold please review code around synchronize blocks. I couldn't come up with something less blocking :(

tsiq-karold:
> @tsiq-karold please review code around synchronize blocks. I couldn't come up with something less blocking :(

GraphQL has the same problem in theory.  I solved it there with a Guava `Suppliers.memoize`.  If we want to avoid the external dependency, we could implement a similar pattern ourselves.  The main difference is they maintain additional state to indicate initialized or not and use double check pattern to synchronize only if not initialised.

Issue #773, opened 2021-01-13
add option to add graphQL operation name to the URL


tsiq-clemens:
All graphQL URLs look alike: `/graphql`. It is a big help to have the distinctive `operationName` in the URL as a query param, turning the URL into something like this: `/graphql?operation=doMyLaundry` which would show up in steps, logs, etc. and add tremendous value for developers (I have first-hand experienced with and without this param).

The `operation` query parameter is ignored by all implementations I know, but we can make it optional, e.g. offer a config param `graphQLDoNotShowOperation` or some such.

MykolaGolubyev:
I believe @tsiq-karold created a similar #583 issue. 
Steps is a good point. Even if we have a dedicated graphql tab,  steps would still be lacking. But I think #605 should address those. 
I can help with report one since it will take longer for you to get to the whole dev setup. 
And you can prioritize #605 then.
How does it sound?

tsiq-karold:
Those issues you linked are about improving visibility/debugging *within* webtau.  Appending it to the URL will also help with server side visibility/debugging where logs, for example, often will have URLs but not bodies.

MykolaGolubyev:
I see. We have HttpConigurations that can update url at runtime. Maybe you can use that one? Or implement similar one for GraphQL?

tsiq-clemens:
Can a `HttpConfiguration` help us here, since the operation comes from the `GraphQLRequest`?

MykolaGolubyev:
`HttpConfiguration` will still be called for graphql calls. The problem is probably going to be that at that time you won't know the operation name unless you store it somewhere in thread local context. 
But if we introduce `GraphQLHttpConfiguration` or something like this that will have an extra context, than you could amend each request as you see fit. 

At this point I think that adding `operation=myOperation` to core logic of every graphql call could be confusing to users. Hence proposing to figure out an extension point that you can leverage. 

It could be a switch flag also in a config. 

To double check, if two issues mentioned above are resolved, will you still benefit from adding `operation` param to each call?

tsiq-karold:
> To double check, if two issues mentioned above are resolved, will you still benefit from adding operation param to each call?

Yes, we will because those two issues only help within the context of webtau. The main benefit of appending to the url is to help debugging server side.

I agree that making the url construction pluggable (directly or indirectly) and is a better option than having the appending be a core feature of webtau. I don't think many users would be interested in that specifically.

MykolaGolubyev:
[...]

PR #809, opened 2021-02-10
graphql: custom GraphQLHttpConfiguration

tsiq-clemens:
fixes https://github.com/testingisdocumenting/webtau/issues/773

### DIFF
--- a/webtau-docs/znai/GraphQL/customized-graphql-urls.md
+++ b/webtau-docs/znai/GraphQL/customized-graphql-urls.md
@@ -1,8 +1,8 @@
 # Customizing the GraphQL Endpoint
 
-By default, WebTau assumes that all requests are made to a `/graphql` endpoint. 
+By default, webtau assumes that all requests are made to a `/graphql` endpoint. 
 In some cases it may make sense to customize this URL, e.g. to hit a non-standard endpoint or
-to decorate the query with an ignored query parameter that contains the query's `operationName`. Execution logs as well as WebTau's step reports will then become easier to read and debug, especially since the operation's name is in the request's payload and is usually not logged by request loggers.
+to decorate the query with an ignored query parameter that contains the query's `operationName`. Execution logs as well as webtau's step reports will then become easier to read and debug, especially since the operation's name is in the request's payload and is usually not logged by request loggers.
 
 In order to change the URL, you need to implement a `GraphQLHttpConfiguration` and ensure it gets loaded at runtime via the Java ServiceLoader.
 
--- a/webtau-docs/znai/GraphQL/customized-graphql-urls.md
+++ b/webtau-docs/znai/GraphQL/customized-graphql-urls.md
@@ -1,0 +1,20 @@
+# Customizing the GraphQL Endpoint
+
+By default, Webtau assumes that all requests are made to a `/graphql` endpoint. 
+In some cases it may make sense to customize this URL, e.g. to hit a non-standard endpoint or
+to decorate the query with an ignored query parameter that contains the query's `operationName`. Execution logs as well as Webtau's step reports will then become easier to read and debug, especially since the operation's name is in the request's payload and is usually not logged by request loggers.
+
+In order to change the URL, you need to implement a `GraphQLHttpConfiguration` and ensure it gets loaded at runtime via the Java ServiceLoader.
+
+Here is an example for such a `GraphQLHttpConfiguration` in groovy:
+```groovy
+private withOperationAsQueryParam = new GraphQLHttpConfiguration() {
+    @Override
+    String requestUrl(String url, GraphQLRequest graphQLRequest) {
+        if (null != graphQLRequest.operationName && !graphQLRequest.operationName.isEmpty()) {
+            return "/my-custom-graphql-endpoint?operation=${graphQLRequest.operationName}"
+        }
+        return url
+    }
+}
+```
--- a/webtau-docs/znai/GraphQL/customized-graphql-urls.md
+++ b/webtau-docs/znai/GraphQL/customized-graphql-urls.md
@@ -1,20 +1,11 @@
 # Customizing the GraphQL Endpoint
 
-By default, Webtau assumes that all requests are made to a `/graphql` endpoint. 
+By default, WebTau assumes that all requests are made to a `/graphql` endpoint. 
 In some cases it may make sense to customize this URL, e.g. to hit a non-standard endpoint or
-to decorate the query with an ignored query parameter that contains the query's `operationName`. Execution logs as well as Webtau's step reports will then become easier to read and debug, especially since the operation's name is in the request's payload and is usually not logged by request loggers.
+to decorate the query with an ignored query parameter that contains the query's `operationName`. Execution logs as well as WebTau's step reports will then become easier to read and debug, especially since the operation's name is in the request's payload and is usually not logged by request loggers.
 
 In order to change the URL, you need to implement a `GraphQLHttpConfiguration` and ensure it gets loaded at runtime via the Java ServiceLoader.
 
-Here is an example for such a `GraphQLHttpConfiguration` in groovy:
-```groovy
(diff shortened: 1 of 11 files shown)


webtau-docs/znai/GraphQL/customized-graphql-urls.md
```diff
@@ -0,0 +1,20 @@
+# Customizing the GraphQL Endpoint
+
+By default, Webtau assumes that all requests are made to a `/graphql` endpoint.
```
tsiq-karold, line 3: We should standardise on webtau vs. Webtau vs. WebTau.  Pretty sure I've seen all three in our docs 😁

webtau-docs/znai/GraphQL/customized-graphql-urls.md
```diff
@@ -0,0 +1,20 @@
...
+
+By default, Webtau assumes that all requests are made to a `/graphql` endpoint. 
+In some cases it may make sense to customize this URL, e.g. to hit a non-standard endpoint or
+to decorate the query with an ignored query parameter that contains the query's `operationName`. Execution logs as well as Webtau's step reports will then become easier to read and debug, especially since the operation's name is in the request's payload and is usually not logged by request loggers.
+
+In order to change the URL, you need to implement a `GraphQLHttpConfiguration` and ensure it gets loaded at runtime via the Java ServiceLoader.
+
+Here is an example for such a `GraphQLHttpConfiguration` in groovy:
+```groovy
+private withOperationAsQueryParam = new GraphQLHttpConfiguration() {
```
tsiq-karold, line 11: Should this be pulled in from some test code?

webtau-test-server/src/main/java/org/testingisdocumenting/webtau/http/testserver/GraphQLResponseHandler.java
```diff
@@ -59,7 +59,7 @@ public GraphQLResponseHandler(GraphQLSchema schema, Handler additionalHandler) {
     @Override
     public void handle(String url, Request baseRequest, HttpServletRequest request,
                        HttpServletResponse response) throws IOException, ServletException {
-        if ("/graphql".equals(baseRequest.getOriginalURI())) {
+        if (baseRequest.getOriginalURI().startsWith("/graphql")) {
```
tsiq-karold, line 62: Hmm, this doesn't play so nicely with the `GraphQLHttpConfiguration` concept because the URL may not be `/graphql`.  Perhaps we drop this check and rely purely on POST + validation of the request + response?

tsiq-karold (commented):
Concept looks good to me.  A couple small points about the impl though.

webtau-docs/znai/GraphQL/customized-graphql-urls.md
tsiq-clemens: Very true. 
The master pom says `<description>WebTau - web test automation. REST/GraphQL/DB/CLI/UI testing framework.</description>` , so maybe we go with that? Up to @MykolaGolubyev 

I can do a global search&replace in .md files in a separate PR.

webtau-docs/znai/GraphQL/customized-graphql-urls.md
tsiq-clemens: Makes sense. I can pull it out of the test code into its own file, yeah.

webtau-test-server/src/main/java/org/testingisdocumenting/webtau/http/testserver/GraphQLResponseHandler.java
tsiq-clemens: it's tricky. The concept of `if (isGraphQLRequest) handleGraphQL() else if (useOtherHandler) doThat else 404` does make sense, especially since this is to support testing of both main handlers.
We could do some request introspection to see whether this is a valid graphql request. Basically most code that you already have in `handleGraphQLPathRequest` , just without auth check and handle call.

webtau-docs/znai/GraphQL/customized-graphql-urls.md
MykolaGolubyev: Fair point. The reason for PascalCase in description is to get the abbreviation across. After that I say we use `webtau` everywhere. Maybe even replace the WebTau in description. 
Would love a PR with this fix!

webtau-docs/znai/GraphQL/customized-graphql-urls.md
MykolaGolubyev: Yeah, take a look at other feature tests, they have `extractSnippet` portions that may help

webtau-graphql-groovy/src/test/groovy/org/testingisdocumenting/webtau/graphql/GraphQLHttpConfigurationsTest.groovy
```diff
@@ -0,0 +1,84 @@
...
+            urls << url
+            return url
+        }
+
+        @Override
+        HttpHeader fullHeader(String fullUrl, String passedUrl, HttpHeader given) {
+            return given
+        }
+    }
+    private GraphQLHttpConfiguration withOperationAsQueryParam = new CustomGraphQLHttpConfiguration()
```
MykolaGolubyev, line 46: please add empty line above. One day we will auto check formatting. Maybe in 2025

MykolaGolubyev (commented):
Is there going to be a shortcut option like `graphqlIncludeOperationName = true`?

tsiq-clemens:
There's no shortcut option at the moment, but we can discuss how to use the new extension point. 
We could pull the sample implementation up to src/main, and add it to `GraphQLHttpOperations` if `graphqlIncludeOperationName` is set. That would free users (like me) from writing and ServiceLoading our own version.
You like?

webtau-test-server/src/main/java/org/testingisdocumenting/webtau/http/testserver/GraphQLResponseHandler.java
tsiq-clemens: thinking about this some more, is there anything that can tell us with near certainty that a request is a graphql request , outside of the endpoint? Only `query` is required, `operationName` and `variables` are optional.
Elastic search requests have a `query`, too, and possibly many custom endpoints.
Since this is a test server, it might be ok requiring the endpoint to at least start with `/graphql`, although it's admittedly not watertight anymore after this PR but hits 99% of our use cases.

tsiq-clemens:
we can even offer 2 config options for a) overriding `graphqlEndpoint` and b) `graphqlIncludeOperationName` that are both handled by this simple `GraphQLHttpOperation`

MykolaGolubyev:
> we can even offer 2 config options for a) overriding `graphqlEndpoint` and b) `graphqlIncludeOperationName` that are both handled by this simple `GraphQLHttpOperation`

separate options look good to me. Feel free to add `graphqlIncludeOperationName` in a separate PR

webtau-test-server/src/main/java/org/testingisdocumenting/webtau/http/testserver/GraphQLResponseHandler.java
tsiq-karold: Sorry about the delay here.

In terms of detecting it's GraphQL I think looking for `query` and then trying to parse it as a GraphQL request is as good as you're going to get.  ES and others are highly unlikely to use GraphQL syntax in a `query` variable.

That said, given that hardly anyone will change from `/graphql` what you have is probably ok for the time being.  Might be worth calling this out in docs though.

tsiq-karold:
> we can even offer 2 config options for a) overriding `graphqlEndpoint` and b) `graphqlIncludeOperationName` that are both handled by this simple `GraphQLHttpOperation`

I would rename the `graphqlIncludeOperationName` option because it's not at all clear this ends up appending a query param.

tsiq-karold (approved):

PR #824, opened 2021-02-17
core: cleanup registration to be part of teardown 

MykolaGolubyev:


MykolaGolubyev:
@tsiq-karold could you please land a pair of eyes here?

MykolaGolubyev:
@tsiq-clemens could you please land a pair of eyes here as well?

PR #834, opened 2021-02-25
build: git: ignore runtime-generated test files

tsiq-clemens:
dealing with a few files that tend to hang around and confuse me when I prep commits

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@ webtau.*report*.html
 webtau-report-*.html
 *-webtau-report.html
 webtau.graphql-report.json
-run-details.actual.json
 .webtau.cache.json
 .vertx
 todo.md
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,8 @@ node_modules
 webtau.*report*.html
 webtau-report-*.html
 *-webtau-report.html
+webtau.graphql-report.json
+run-details.actual.json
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -7,6 +7,8 @@ node_modules
 webtau.*report*.html
 webtau-report-*.html
 *-webtau-report.html
+webtau.graphql-report.json
+run-details.actual.json
```
MykolaGolubyev, line 11: this should go away when test is passes, so maybe a good idea to have it not ignored. dunno.

MykolaGolubyev:
I would rename PR to `git: `. `core: ` is for actual core logic

.gitignore
tsiq-clemens: you don't want to add it to git, though, right?

MykolaGolubyev: 50/50. Add it. No harm. Was thinking could be a good signal to remind you to rerun a test

tsiq-clemens: it is more prominent, I give you that. I'll remove it from here.

MykolaGolubyev (approved):

tsiq-karold:
> I would rename PR to `git: `. `core: ` is for actual core logic

I think I've used `build:` for this sort of thing, bit broader than git.

PR #842, opened 2021-03-01
allow nesting of the already active persona

### LINKED ISSUE #835: personas: nesting the same persona should be allowed
I think we should allow this:
```groovy
def john = Persona('John')

john {
  john {
    http.post(...)
  }
}
```

Obviously nesting that explicitly is pointless.  However, some utility functions could well operate in the context of an explicit persona and their calls wrap calls to that function with the same persona for double checking and making things more obvious.

Happy to discuss this here prior to implementing though :)

tsiq-karold:
fixes #835


MykolaGolubyev (approved):
Nice

PR #845, opened 2021-03-04
matchers: set and table data compareTo handler fix

### LINKED ISSUE #156: Set properties cannot be compared with TableData
Example:
Set aSet = myBean.getChildrenSet()
aSet.should == [ ... table data ...] // throws exception that Set cannot be converted to a List.
aSet.toList().should == [ ... table data ...] // works fine

This prevents sub collection matching if I want to verify properties of myBean, e.g.
myBean.should == [ propA: 'expectedA', childrenSet: expectedChildrenTableData ]

MykolaGolubyev:
closes #156


tsiq-clemens (approved):
I like this, especially matching by key column

PR #846, opened 2021-03-04
http: capture start/elapsed of failed http call

### LINKED ISSUE #844: requests that time out report incorrect start time
webtau-timeout

* one problem is that after upgrading to webtau-1.38 (defaulting the http connection's `connectTimeout` and `readTimeout` to 30s), we see SocketTimeout exceptions that we did not get before. The corresponding requests on the backend appear to have finished in milliseconds to a few seconds, whereas webtau reports the scenario as >30s
* steps that show a timeout are reported with a `19:00:00.000` start time, which probably means that its `startTime` was never set to the current time, meaning `startClock()` never happened.

`Http`'s `executeAndValidateHttpCall` has a try-finally clause, but no catch, is that part of the problem?

```
        try {
            return step.execute(StepReportOptions.REPORT_ALL);
        } finally {
            lastValidationResult.set(validationResult);
            step.addPayload(validationResult);
        }
```

MykolaGolubyev:
closes #844

### DIFF
--- a/webtau-core/src/main/java/org/testingisdocumenting/webtau/time/Time.java
+++ b/webtau-core/src/main/java/org/testingisdocumenting/webtau/time/Time.java
@@ -30,4 +30,14 @@ public class Time {
     public static void setTimeProvider(TimeProvider replacement) {
         timeProvider.set(replacement != null ? replacement : systemTimeProvider);
     }
+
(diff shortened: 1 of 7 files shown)


tsiq-clemens (approved):
LGTM

PR #858, opened 2021-03-17
db: lazy primary datasource init

MykolaGolubyev:
@tsiq-karold please review.
I will extract lazy-init common one in a separate PR and will use in other places


tsiq-karold (approved):
Looks good, makes sense to extract out and use elsewhere.  Will hopefully get us a step closer to removing the guava dependency :)

PR #865, opened 2021-03-21
core: add repeatStep to streamline reporting

MykolaGolubyev:
@tsiq-karold @tsiq-clemens this is a step forward towards step retries and better performance related testing tooling.


MykolaGolubyev:
@tsiq-karold I looked at graphql code, it shouldn't affect what you have, plus you don't use `repeatStep` yet.
But the idea is to skip 1k+ calls and not store them in report, but still be able to calc performance metrics. 
I will be implementing this for HTTP calls this week, and I think it should be common/similar for graphql

PR #1435, opened 2023-04-17
core: make HandlerMessages public 

vmpn:
... so that developers can reuse the reporting logic in their custom handlers


MykolaGolubyev (approved):
Thank you!

